PR #2, opened 2024-08-21
Onestop scanning implementation

jbwong05:
Draft PR that contains the initial implementation for scanning the Onestop for daily tabs and lines to process within those daily active tabs. This PR is admittedly pretty big and there is definitely some more work to be done with regards to efficiency and cleanup, however it is functional at the moment.

Publicly exposed functions:
* `getAllSpreadsheetTabs(): Sheet[]`
* `getActiveDailyTabs(spreadsheetTabs: Sheet[]): Sheet[]`
* `getRowsWithEvents(spreadsheetTab: Sheet): Row[]`

Added types:
* `Spreadsheet`
* `Sheet`
* `Text`
* `Row`

### DIFF
--- a/src/main/scan.ts
+++ b/src/main/scan.ts
@@ -48,7 +48,9 @@ export function getAllSpreadsheetTabs(): Sheet[] {
 export function getActiveDailyTabs(spreadsheetTabs: Sheet[]): Sheet[] {
     const dailyActiveTabs: Sheet[] = [];
     for(const tab of spreadsheetTabs) {
-        checkIfActiveDailyTab(tab, dailyActiveTabs);
+        if(isActiveDailyTab(tab)) {
+            dailyActiveTabs.push(tab);
+        }
     }
 
     return dailyActiveTabs;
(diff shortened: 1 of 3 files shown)


src/main/scan.ts
```diff
@@ -0,0 +1,285 @@
...
+    return rows;
+}
+
+/**
+ * Checks if a given tab is not hidden and is a daily tab. If it is, add it to the dailyActiveTabs output array
+ * 
+ * @param tab the tab to check
+ * @param dailyActiveTabs output array to add the tab to if it meets the criteria of not being hidden and being a daily tab
+ */
+function checkIfActiveDailyTab(tab: Sheet, dailyActiveTabs: Sheet[]): void {
```
andrewc5716, line 83: in general let's try to avoid output arguments; it makes readability a bit harder. in this case it should be an easier fix since your return type currently is void, so it can be replaced by just doing a boolean check if it's an active daily tab

jbwong05: Output parameter removed. Function was updated to just return a boolean representing whether the given tab is an active and daily tab

andrewc5716 (approved):

PR #3, opened 2024-08-27
add basecamp apis

andrewc5716:
- adds basecamp api calls as functions
- readme significantly updated to add oauth and basecamp integration instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,4 +22,13 @@ npm run build
 npm run deploy
 ```
 
+Register an application:
+Name of application: name it
+Company name:
+URL:
+Products:
+Redirect URI: 
 
+Services > Google Sheets
+
+in test app in bc, need to set redirect uri for each person!?
--- a/README.md
+++ b/README.md
@@ -1,45 +1,54 @@
-# README
+# SD Onestop to Basecamp Integration
+Takes events from the Onestop's daily tabs and populates Basecamp with appropriate todos for each person.
 
-```shell
-npm i
-```
-
-Enable the Google Apps Script API at https://script.google.com/home/usersettings:
-
+## Developer setup
+### Local development environment
+1.  ```npm i ```
+2. Enable the Google Apps Script API at https://script.google.com/home/usersettings:
 ![Enable Apps Script API](https://user-images.githubusercontent.com/744973/54870967-a9135780-4d6a-11e9-991c-9f57a508bdf0.gif)
+3. Download clasp: ```npm install -g @google/clasp```
+4.  ```clasp login ```
 
-```shell
-clasp login
+### Personal Onestop and app script setup
+5. Make your own copy of the Onestop from [this folder](https://drive.google.com/drive/folders/1s_u-hmstlLL1JqRKyKJ7r2AMyyTCCiut) and add your name to it.
+6. In your new sheet, on the menu click Extensions > App Script
+7. Select "Create a new project"
+8. Name it appropriately, e.g. "Onestop Basecamp Integration - My Name"
+9. In the Apps Scripts project, open "Project Settings" on the left bar
+10. Copy your Script ID.
+11. Replace `scriptId` in [`.clasp.json`](./.clasp.json) with your created script ID.
+
+### Personal Basecamp integration setup
+12. Login to [Basecamp](https://launchpad.37signals.com/signin) with your A2N Google account. Ensure you have access to the Acts2Network project before continuing.
+13. Open the [Basecamp Integrations page](https://launchpad.37signals.com/integrations)
+14. Click "Register an application"
+15. Fill out application details:
+- Name it appropriately (e.g. Onestop Integration - My Name)
+- Company and website URL don't matter
+- Select Basecamp 4 under Products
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -1,36 +1,54 @@
...
+- Name it appropriately (e.g. Onestop Integration - My Name)
+- Company and website URL don't matter
+- Select Basecamp 4 under Products
+- For OAuth2 redirect URI, fill in the URI with the following, and replace `YOUR_SCRIPT_ID` with your Google Apps Scripts ID from the earlier section: `https://script.google.com/macros/d/YOUR_SCRIPT_ID/usercallback`. 
+16. Click "Register this app"
+17. Open your app and copy your Client ID and Client Secret for the following step.
+19. Create a new file `environmentVariables.ts` inside the [config](./config/) folder, with the following contents:
+```js
+export const BASECAMP_CLIENT_ID = 'REPLACE_WITH_YOUR_BASECAMP_CLIENT_ID';
+export const BASECAMP_CLIENT_SECRET = 'REPLACE_WITH_YOUR_BASECAMP_CLIENT_SECRET';
```
jbwong05, line 35: Not a blocking review comment, just a question. Is this just temporary or is the plan to hardcode these values?

README.md
```diff
@@ -1,36 +1,54 @@
...
 ```shell
 npm run build
 npm run deploy
 ```
+This will deploy your code to your personal onestop sheet and app script, and use your personal Basecamp integration app.
 
+To run functions, make sure you first open your Google sheet (onestop copy), then open your script from Extensions > App Script. Keep the Google sheet open in a separate tab, as it will be necessary for any Spreadsheet UI interactions. In the App Script code editor, find your function from the dropdown and hit run. If your function makes use of the Spreadsheet UI API, you will need to click back to your Google sheet tab for the interaction.
 
+#### Testing Basecamp API's
+You will need an OAuth2 access token, which you can receive a link for when making any Basecamp request when unauthorized. The easiest way is to call the `checkAuthorization` function from the Google Apps Scripts editor, click back to your Google sheet tab, then follow the instruction in the alert dialog (open the auth link in a new tab, close it, then re-run the function).
```
jbwong05, line 54: Not a blocking review comment, just a question. Does this access token need to be stored somewhere?

src/main/basecamp.ts
```diff
@@ -0,0 +1,150 @@
+import { BASECAMP_CLIENT_ID, BASECAMP_CLIENT_SECRET } from "../../config/environmentVariables";
+
+const BASECAMP_AUTH_URL = 'https://launchpad.37signals.com/authorization/new?type=web_server';
+const BASECAMP_TOKEN_URL = 'https://launchpad.37signals.com/authorization/token?type=web_server';
+const BASECAMP_AUTH_CHECK_URL = 'https://launchpad.37signals.com/authorization.json';
+const OAUTH_BASECAMP_SERVICE_NAME = 'Basecamp';
+const OAUTH_CALLBACK_FUNCTION_NAME = 'oauthCallback'
+const BASECAMP_UNAUTH_ERROR_MSG = 'Basecamp not authenticated. Please try again.';
+const AUTH_DIALOG_MSG = 'You are disconnected from Basecamp. Copy the following link on a new tab to be authorized:\n\n';
+const AUTH_SUCCESS_HTML = 'Connection with Basecamp was successful! You can can close this tab and re-run the program.';
```
jbwong05, line 10: Nit: Word `can` is duplicated

src/main/basecamp.ts
```diff
@@ -0,0 +1,150 @@
...
+const HEADER_USER_AGENT_NAME = 'Google App Script Onestop Basecamp Integration';
+const HEADER_CONTENT_TYPE = 'Content-Type';
+const HEADER_JSON_CONTENT_TYPE = 'application/json';
+
+const HTTP_POST_METHOD = 'post';
+const HTTP_PUT_METHOD = 'put'
+const HTTP_GET_METHOD = 'get'
+
+export function sendBasecampPostRequest(requestUrl: string, requestPayload: Record<string, any>): Record<string, any> {
+    const response: GoogleAppsScript.URL_Fetch.HTTPResponse = UrlFetchApp.fetch(requestUrl, {
```
jbwong05, line 26: Not required but we can use a type alias like `type HTTPResponse = GoogleAppsScript.URL_Fetch.HTTPResponse` at the beginning of this file so the `HTTPResponse` type can be used for better readability

jbwong05 (approved):
Just some minor questions and Nit picks but overall looks good ✅

README.md
andrewc5716: i haven't fully thought it through yet, we'll have a more solid approach in a future task. but here's the factors i'm considering:
- we should avoid putting secrets in version control
- for us to all develop independently, each of ours will need to be separate, hence this is local for each of us and in gitignore
- but we need to figure out how to do this for "prod", e.g. the real onestop

README.md
andrewc5716: it's automatically stored as part of the oauth2 library, i believe in property service. looks like the access token usually lasts for a few weeks before it needs to be renewed

src/main/basecamp.ts
andrewc5716: nice catch

src/main/basecamp.ts
andrewc5716: nice didn't know i could do that, made the change

jbwong05 (approved):
🚀

PR #4, opened 2024-08-30
Add row unique identifiers to metadata

jbwong05:
## Description
Adds the ability to add unique identifiers to rows via [metadata](https://developers.google.com/sheets/api/guides/metadata) as part of https://3.basecamp.com/4474129/buckets/38736474/todos/772966216. Metadata for google sheets allows for a key and a value to be associated with each row independently (the key and value are not tied to each other like they are in a traditional map).
When the scanning of the Onestop rows occurs, a Metadata object is created for each valid event row by setting the key (this causes Google Sheets to create the Metadata object). The value of the Metadata object contains the unique id of the Row. If it is `null`, this indicates it's unique has not yet been assigned indicating the row is new. These unique ids are tied to the specific row and will stay with the row even if the row is moved around in the spreadsheet tab.

## Changes
* Re-added `ts-loader` dependency that was removed as part of #3 as it is required to compile TS with webpack
* Added functions that allow checking, assigning, and retrieving unique ids for rows via metadata
* Adds a metadata field to the Row interface
* Removes a function with an output parameter which we are trying to avoid (added as part of #2 )

### DIFF
--- a/src/main/metadata.ts
+++ b/src/main/metadata.ts
@@ -27,8 +27,15 @@ export function getMetadata(range: Range): Metadata {
  * @param row the Row to retrieve the unique id for
  * @returns unique id for the row; null if one has not been set
  */
-export function getRowUniqueId(row: Row): string | null {
-    return row.metadata.getValue();
+export function getId(row: Row): string {
+    const id = row.metadata.getValue();
+
+    if(id === null) {
+        throw Error(`Row does not have an id: [${row.startTime}, ${row.endTime}, ${row.who}, 
+            ${row.what.value}, ${row.where.value}, ${row.inCharge.value}, ${row.helpers.value}]`);
+    }
+
+    return id;
 }
 
 /**
@@ -37,7 +44,7 @@ export function getRowUniqueId(row: Row): string | null {
  * @param row the row to assign a new unique id to
  * @returns the new unique id
  */
-export function assignRowUniqueId(row: Row): string {
+export function assignId(row: Row): string {
     const newId: string = Utilities.getUuid();
     row.metadata.setValue(newId);
     return newId;
@@ -49,6 +56,6 @@ export function assignRowUniqueId(row: Row): string {
  * @param row the row to check the unique id of
  * @returns boolean representing whether a given Row has been assigned a unique id or not
  */
-export function rowHasUniqueId(row: Row): boolean {
-    return getRowUniqueId(row) !== null;
+export function hasId(row: Row): boolean {
+    return row.metadata.getValue() !== null;
 }
--- a/src/main/metadata.ts
+++ b/src/main/metadata.ts
@@ -1,0 +1,54 @@
+const ROW_ID_KEY: string = "rowId";
+
+/**
+ * Retrieves the metadata object for a given range. If the metadata object does not exist,
+ * this function will set the row id key so the metadata object is created and a reference
+ * can be held
+ * 
+ * @param range Range representing an event row
+ * @returns reference to the Metadata object for this row
+ */
+export function getMetadata(range: Range): Metadata {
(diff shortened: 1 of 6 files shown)


src/main/metadata.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+ * Retrieves the unique id for a Row. The unique id is stored in the value of the metadata.
+ * If the unique id has not been set, null will be returned
+ * 
+ * @param row the Row to retrieve the unique id for
+ * @returns unique id for the row; null if one has not been set
+ */
+export function getRowUniqueId(row: Row): string | null {
```
andrewc5716, line 30: non-blocking comment: in my mind "id" is usually unique, so i'm wondering if dropping "unique" from these function names would help with conciseness. not 100% sure of this suggestion but if we were to also infer we're working with rows b/c of the input argument, we could even call them "getId", "generateId", "hasId" etc.

src/main/scan.ts
```diff
@@ -70,7 +71,15 @@ export function getRowsWithEvents(spreadsheetTab: Sheet): Row[] {
     const rows: Row[] = [];
     for(let i = 0; i < cellData.length; i++) {
         const rowData: CellData[] = cellData[i];
-        processRow(rowData, rows, currentDate);
+        if (isValidEventRow(rowData)) {
+            // Google sheets is 1 indexed
```
andrewc5716, line 75: sick lol

andrewc5716 (dismissed):

src/main/metadata.ts
andrewc5716, line 30: as far as possible, we should try to avoid returning null - it's just waiting for a NPE down the road. i'd rather throw an exception immediately here and we can identify the source of the issue in a controlled manner, vs the NPE popping up somewhere else and it may take longer to debug why

andrewc5716: i see it's used for rowHasUniqueId, but that could just use row.metadata.getValue() !== null it seems.
i don't want to block this PR since i'll be gone over the weekend, so you can feel free to address this in the next PR or get someone else's approval

jbwong05: Will drop Row and Unique from the function names

src/main/scan.ts
jbwong05: lol

src/main/metadata.ts
jbwong05: I'll address this in this PR by throwing an exception. Will get a review from someone else

src/main/row.ts
```diff
@@ -0,0 +1,72 @@
...
+    return row.metadata.getValue() !== null;
+}
+
+/**
+ * Returns a string representation of the given row
+ * 
+ * @param row the row to return a string representation for
+ * @returns string representation of the given row
+ */
+function toString(row: Row): string {
```
andrewc5716, line 68: this makes me think for the future it might be helpful to add the sheet tab name and maybe the row number. out of scope for this pr, but we might want to consider adding it later when it makes sense

src/main/row.ts
```diff
@@ -0,0 +1,72 @@
...
+    return id;
+}
+
+/**
+ * Assigns a new unique id to a row
+ * 
+ * @param row the row to assign a new unique id to
+ * @returns the new unique id
+ */
+export function assignId(row: Row): string {
```
andrewc5716, line 46: nitpick, "assignId" makes me think we're going to pass in an id to assign. i'm wondering if something like "generateId" would be more clear, or if we even need to go more explicit like "generateIdForRow()"

jbwong05: Renamed to `generateIdForRow()`

andrewc5716 (approved):

jkzang (approved):

PR #5, opened 2024-09-01
Local DB

Isaac-Otero:
Creates a local DB, used mock data but will create a local DB using properties service


Isaac-Otero:
Will PR when it just works

PR #7, opened 2024-09-09
basecamp todos

andrewc5716:
Adds functions for creating and updating todos in Basecamp.

Note that to create a todo it requires `TodolistIdentifier`, which takes a project id and todolist id. These will need to be hardcoded for our first phase (so it will always create todos in a specific project and specific todolist). In the future this can be more flexible to put tasks in a ministry domain specific project.

Additional changes:
- Added types `JsonData`, which can be either `JsonObject` or `JsonArray`. This is more readable than `Record<string, any>`. Also, this forced some safety with only paginating on array responses (for people.json).


src/main/basecamp.ts
```diff
@@ -22,9 +22,30 @@ const HTTP_POST_METHOD = 'post';
 const HTTP_PUT_METHOD = 'put'
 const HTTP_GET_METHOD = 'get'
 
-type HTTPResponse = GoogleAppsScript.URL_Fetch.HTTPResponse;
+const BASECAMP_API_URL = 'https://3.basecampapi.com';
+const A2N_BASECAMP_ORG_ID = '4474129';
```
jbwong05, line 26: For the long term will this remain hardcoded or will this be moved to be with the rest of our config values?

jbwong05 (approved):
:rocket:

src/main/basecamp.ts
andrewc5716: good callout. for the medium term and possibly long term this could be hardcoded since the organization will always be A2N. but it could live in config values too, though it'd always stay constant. when we flesh out our environment specific configs more, we can see what makes the most sense

PR #8, opened 2024-09-15
Populate People in Local Db

jbwong05:
## Description
Adds a function which hits the `/people.json` API, filters the results with a hardcoded list of people on SD team from the Master Rotation, and then adds a mapping of name -> id to the script properties Properties Service as well as a flag indicating that the people name to id map was populated

## Need help testing
This PR is in draft because I will need some help testing this functionality. When I hit the `people.json` API, I get a response of only the 4 people working on this project and not other people that are part of the A2N wider org


src/main/people.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    // Reduce all of the people to a single map
+    const personNameIdMap: PersonNameIdMap = sdPeople.reduce((map, person) => {
+        map[person.name] = person.id;
+        return map;
+    }, {} as PersonNameIdMap);
+
+    personNameIdMap[PEOPLE_POPULATED_KEY] = "true";
+
+    setScriptProperties(personNameIdMap);
```
andrewc5716, line 30: i just realized, one downside of saving a map of personName -> personId is that when ppl are removed from the sd list, we'd have to manually call a function to clean up that specific person from properties service otherwise it'll remain there forever. similar issue with removing rowIds too i think.
instead, what do you think of storing the whole people map/object as the value, and the key would be a constant like "PEOPLE_MAP"? then we could more easily do a full replacement of the people map object and not worry about losing any ppl references. the downside here is if we were to retrieve the whole people map object each time it'd be slow, but we could cache it in memory by retrieving the full object once. what are your thoughts?

jbwong05: Script properties actually have a UI that allows you to delete individual values, but I am ok storing the whole map/object and using an in memory cache. Makes the script properties a lot cleaner especially if we are going to store under values here too. I don't think the in memory cache would persist between different triggers but within one trigger we would only need to fetch the map once

jbwong05:
Changes made as per suggestions made during PR review:
* Switched to storing the entire people map object with an in memory cache
* Fix Error class exports/imports
* Switched to the `/projects/{PROJECT_ID}/people.json` api due to issues with the `/people.json` api
  * The PROJECT_ID is currently hardcoded to our SD Basecamp Integration id for testing. We can consider moving this config value to script properties in the future

src/main/people.ts
jbwong05: Should we also consider doing something similar for the row mapping object as well?

src/main/people.ts
```diff
@@ -0,0 +1,99 @@
...
+    // Check the in memory cache first
+    if(cachedPersonNameIdMap !== null) {
+        return getPersonIdFromCache(personName);
+    }
+
+    // Otherwise attempt to fetch it from the script properties
+    const personNameIdMap: string | null = getScriptProperty(PEOPLE_MAP_KEY);
+
+    if(personNameIdMap === null) {
+        throw new PeopleNotPopulatedError("People have not been populated in the local db");
```
andrewc5716, line 60: what do you think about just making the call to populatePeopleInDb here? that way from the external caller, they just call getPersonId and don't need to remember to populate it first.

src/main/people.ts
```diff
@@ -0,0 +1,99 @@
...
+type PersonNameIdMap = {[key: string]: string};
+
+// Project ID hardcoded to our SD Basecamp Integration project for testing
+// In the future we may consider moving this to a script property as a config value
+const PROJECT_ID: string = "38736474";
+const PEOPLE_PATH: string = `/projects/${PROJECT_ID}/people.json`;
+const PEOPLE_MAP_KEY: string = "PEOPLE_MAP";
+
+// Regex to match a person's name followed by a city. ex. John Doe (SD)
+const CITY_REGEX: string = "^(.+?) (.+?) \((.+)\)$";
```
andrewc5716, line 16: i suspect this won't work for filipino ppl like "Jonathan Koby Cayaban (SD)". i think it's safe to just chop off the parentheses portion and everything before that is the name.

src/main/people.ts
jbwong05: I thought about this during implementation and forgot to put it in. Nice catch

src/main/people.ts
jbwong05: Updated the matching to match any string that comes before `()` so we can just chop off the parentheses

src/main/people.ts
```diff
@@ -0,0 +1,99 @@
...
+        populatePeopleInDb();
+        return getPersonIdFromCache(personName);
+    }
+
+    // Populate the cache and fetch the person id if it exists
+    cachedPersonNameIdMap = JSON.parse(personNameIdMap);
+    if(cachedPersonNameIdMap !== null && cachedPersonNameIdMap.hasOwnProperty(personName)) {
+        return cachedPersonNameIdMap[personName];
+    } else {
+        throw new PersonMissingIdError(`Person does not have an id: [personName: ${personName}]`);
```
andrewc5716, line 68: lmk your thoughts; i'm wondering if instead of throwing an error here, we change the function signature to return `string | undefined`, so that the caller would be forced to check, vs needing to know to catch this specific error

src/main/people.ts
```diff
@@ -0,0 +1,99 @@
...
+ * the script properties with a map from person name to Basecamp person id. Can be invoked manually
+ * from the Google Apps Script Developer UI
+ */
+export function populatePeopleInDb(): void {
+    const requestUrl: string = getBasecampUrl() + PEOPLE_PATH;
+    const peopleData: Person[] = sendPaginatedBasecampGetRequest(requestUrl) as Person[];
+
+    // Reduce all of the people to a single map
+    const personNameIdMap: PersonNameIdMap = peopleData.reduce((map: PersonNameIdMap, person: Person) => {
+        const match: RegExpMatchArray | null = person.name.match(CITY_REGEX);
```
andrewc5716, line 30: small nitpick, the extracting of the person's name feels like it should be a separate function to maintain the same level of abstraction within the function

andrewc5716 (dismissed):
lgtm; if you want to address the small comments go ahead, but approving anyway to unblock you if you want to move on and address those later

src/main/people.ts
jbwong05: Yea makes sense to me

andrewc5716 (approved):

PR #9, opened 2024-09-24
Dev/create todo for leads

jbwong05:
This PR is just a cleanup of Jack's work to expose a function that creates a Basecamp Todo Request object from a given row. It probably could use some more work but still submitting it for feedback. The description will probably need to be updated in the future to be more in-line with what Audrey wants.

Note: It looks like the Lead column will also need to be cleaned up (not just the helper column) on the Onestop. Many of the names of people in the Lead column are missing last names especially when it's couples assigned to lead an event/task


andrewc5716 (approved):
actually, lgtm

andrewc5716:
and yes the lead column will need to be cleaned up. during management mtg they'll make sure it's full names and comma delimited, until we make the inputs more lenient in the future

PR #11, opened 2024-10-01
Core logic create todo new row

jbwong05:
## Description
Core logic for creating new Todos
Screenshot 2024-10-01 at 2 52 55 PM

## Changes Included in this PR
* Added missing type declarations (previously missed)
* Fixed a bug where the `hasId()` check for a row was always true
* Core logic for scanning through the Onestop and sending requests to Basecamp to create Todos for event rows
* Utility function to clear all row metadata/document properties (this is useful while debugging and can be triggered manually to "start fresh")
* Updated the row basecamp mapping to include the Todo ids
* Removed the batched `saveRows()` function as we will be looping
* Small refactoring to abstract away more of the event row fetching and basecamp request fetching away from the core logic

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -12,7 +12,7 @@ const DEFAULT_TODOLIST_IDENTIFIER: TodolistIdentifier = {
  * Main entry point for the Onestop to Basecamp Integration that contains the core logic for
  * adding/updating/deleting Basecamp Todos based on the rows on the Onestop
  */
-export function main(): void {
+export function importOnestopToBascamp(): void {
     const eventRows: Row[] = getEventRowsFromSpreadsheet();
     const processedRowIds: string[] = [];
 
@@ -35,7 +35,7 @@ function processExistingRow(row: Row): void {
 
 /**
  * Processes new event rows by creating a new Basecamp Todo for each of the roles associated with that 
- * particular event row. If at least on Todo is creaed in Bascamp, the row is saved for easier retrieval
+ * particular event row. If at least on Todo is created in Bascamp, the row is saved for easier retrieval
  * and updating later
  * 
  * @param row 
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -12,7 +12,7 @@ const DEFAULT_TODOLIST_IDENTIFIER: TodolistIdentifier = {
  * Main entry point for the Onestop to Basecamp Integration that contains the core logic for
  * adding/updating/deleting Basecamp Todos based on the rows on the Onestop
  */
-export function importOnestopToBascamp(): void {
+export function importOnestopToBasecamp(): void {
     const eventRows: Row[] = getEventRowsFromSpreadsheet();
     const processedRowIds: string[] = [];
 
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -1,6 +1,6 @@
 import { PROJECT_ID } from "./basecamp";
-import { generateIdForRow, getBasecampTodoForLeads, getBasecampTodosForHelpers, getId, hasId, saveRow } from "./row";
-import { getActiveDailyTabs, getAllSpreadsheetTabs, getRowsWithEvents } from "./scan";
+import { generateIdForRow, getBasecampTodoRequestsForRow, getId, hasId, saveRow } from "./row";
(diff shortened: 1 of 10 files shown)


src/main/main.ts
```diff
@@ -0,0 +1,70 @@
...
+    deleteOldRows(processedRowIds);
+}
+
+function processExistingRow(row: Row): void {
+    // Will be implemented as part of https://3.basecamp.com/4474129/buckets/38736474/todos/7717428992
+}
+
+/**
+ * Processes new event rows by creating a new Basecamp Todo for each of the roles associated with that 
+ * particular event row. If at least on Todo is creaed in Bascamp, the row is saved for easier retrieval
```
andrewc5716, line 38: small typo "created"

src/main/main.ts
```diff
@@ -0,0 +1,70 @@
...
+ * and updating later
+ * 
+ * @param row 
+ */
+function processNewRow(row: Row): void {
+    const basecampTodoRequests: BasecampTodoRequest[] = getBasecampTodoRequestsForRow(row);
+    const basecampTodoIds: string[] = createNewTodos(basecampTodoRequests);
+
+    if(basecampTodoIds.length > 0) {
```
andrewc5716, line 47: in the case of an execution crash, to me it feels safer to immediately save the row id once the basecamp request succeeds, what do you think? otherwise worst case you have a bunch of orphaned todos in basecamp

src/main/main.ts
```diff
@@ -0,0 +1,70 @@
...
+const DEFAULT_TODOLIST_IDENTIFIER: TodolistIdentifier = {
+    projectId: PROJECT_ID,
+    todolistId: TODOLIST_ID
+};
+
+/**
+ * Main entry point for the Onestop to Basecamp Integration that contains the core logic for
+ * adding/updating/deleting Basecamp Todos based on the rows on the Onestop
+ */
+export function main(): void {
```
andrewc5716, line 15: later we can think of a better name for this - this function will be called both manually from a menu and periodically based on a schedule. importOnestopToBascamp?

src/main/row.ts
```diff
@@ -76,58 +76,28 @@ export function generateIdForRow(row: Row): string {
  * @returns boolean representing whether a given Row has been assigned a unique id or not
  */
 export function hasId(row: Row): boolean {
-    return row.metadata.getValue() !== null;
+    const id: string | null = row.metadata.getValue();
+
+    return id !== null && id !== "";
```
andrewc5716, line 81: nasty, what case does it end up being ""?

src/main/main.ts
andrewc5716: i'm actually ok with main too

andrewc5716 (commented):
i like the clean code here 👍

src/main/row.ts
andrewc5716: if you could add the reasoning in as a comment that may be helpful

jbwong05: Same issue as describe here https://github.com/andrewc5716/onestop-basecamp-integration/blob/9d1ead0d3b674da2bb4cd600ec6ea23df6103ae1/src/main/row.ts#L51-L55
I'll add a comment to make this clear

src/main/main.ts
jbwong05: @andrewc5716 Just for clarification, each row can have multiple Basecamp Todo Requests. Are you saying to save the row after each individual request? 

Another option is we could also wrap the Http request call in a try catch to catch and handle any errors so execution doesn't halt and we can continuing processing as normal saving rows with successful basecamp todo ids so they don't get orphaned. Ideally we would also have some retry mechanism/exponential backoff

Another thing that comes to mind is if one row has 3 Basecamp Todo Requests and request 1 and 2 are successful but request 3 fails. If we continue processing and save the row with ids for request 1 and 2, we have no way of knowing in the future that we need to come back and retry 3 at a later time. We may need a mechanism to handle this

src/main/main.ts
jbwong05: Didn't see the followup comment. Renamed to `importOnestopToBascamp `

jkzang (approved):

PR #12, opened 2024-10-04
Dev/create trigger

jkzang:
This is a pretty simple PR that adds a function that create a new menu on the onestop containing a button for users to manually trigger the importOnestopToBasecamp() function


.clasp.json
```diff
@@ -1,4 +1,4 @@
 {
-  "scriptId": "1N22pZ29WY4DF-KV-JZ7jLh92f44pWrgL0rUqLviBbQ8ECZCEMbpvxi9V",
+  "scriptId": "1wjCZWISevP9KgxxiNM9OPP9Aq3VnGIxVfS6BTj586WhfrKWTg8rRmvgq",
```
andrewc5716, line 2: you probably accidentally committed this, can you remove this from the pr?

src/main/main.ts
```diff
@@ -10,9 +11,13 @@ const DEFAULT_TODOLIST_IDENTIFIER: TodolistIdentifier = {
 
 /**
  * Main entry point for the Onestop to Basecamp Integration that contains the core logic for
- * adding/updating/deleting Basecamp Todos based on the rows on the Onestop
+ * adding/updating/deleting Basecamp Todos based on the rows on the Onestop as well as a function to
+ * create a menu item to give users the ability to manually trigger the import process. 
  */
 export function importOnestopToBasecamp(): void {
+
+    updateManualTriggerMenuUiOnOneStop();
```
andrewc5716, line 19: question - so each time this function is called, the menu needs to be updated?

src/index.ts
```diff
@@ -14,4 +14,5 @@ export * from './main/menu';
 
 // Assign `global` so gas-webpack-plugin will generate a stub for `onOpen` in the top-level scope. A top-level
 // declaration is needed for Google Apps Script to detect the function, so it is available as a trigger.
-global.onOpen = onOpen;
```
andrewc5716, line 17: i suspect onOpen might be something google sheets or google apps scripts actually uses? like when the sheet opens, it'll call this function?

src/main/main.ts
jkzang: Unless we are changing the menu all the time, we probably don't need to update it each time. We can just keep it as a manually  function from the AppScript editor

src/index.ts
jkzang: Yeah that's precisely what it does. When the sheet opens it runs the logic that creates the menu:

```
export const onOpen = (e: OpenEvent) => {
  const tabSyncMenu: GoogleAppsScript.Base.Menu = SpreadsheetApp.getUi()
    .createMenu('OneStop Utils');

  tabSyncMenu.addToUi();
}
```

I guess we could keep it so that the menu gets recreated every time someone opens the OneStop. However, I feel like that would produce even more unecessary executions than when running  `updateManualTriggerMenuUiOnOneStop()` every time `importOnestopToBasecamp()` runs which I now think is also unnecessary

src/main/main.ts
andrewc5716: the menu should be static, at least with our use case. if we write new code that adds a new menu option though, does that mean we'd need to run the function manually to update the menu? ideally we never have to call anything manually and the menu should reflect what we decide in the code

src/index.ts
andrewc5716: so if onOpen isn't called, then no menu is created? or is it just the last menu from the last execution?

i do see your point though; are you able to tell if the sheet loads more slowly as a result? if there's no noticeable difference, it might be ok to have this run on every open, even if unnecessary. but if there's any noticeable slow down, i'm in favor of saving on these extra executions like you're bringing up

src/main/main.ts
andrewc5716: maybe we'd have a function in menu.ts with a comment saying it needs to be run whenever something new is added?

jkzang: If we expect the menu to be mostly static then I think we can take it out from both the onOpen and the importOneStopToBasecamp functions and only run it manually via the AppScript editor. 


> the menu should be static, at least with our use case. if we write new code that adds a new menu option though, does that mean we'd need to run the function manually to update the menu? ideally we never have to call anything manually and the menu should reflect what we decide in the code

If we expect the menu to be mostly static then I think we can take it out from both the onOpen and the importOneStopToBasecamp functions and only run it manually via the AppScript editor.

jkzang: > maybe we'd have a function in menu.ts with a comment saying it needs to be run whenever something new is added?

Sure I can add that comment

.clasp.json
jkzang: Should we just ignore this file?

src/index.ts
jkzang: I'll have to dig into whether onOpen persists. I have a feeling that it doesn't because when I reload my sheet the custom menu is not there but then pops up after 3 seconds

jkzang: Okay, it actually doesn't persist. It needs to get added each time onOpen. I'll change that.

jkzang: @andrewc5716  Okay I think this PR should be good now?

src/main/main.ts
```diff
@@ -10,9 +10,11 @@ const DEFAULT_TODOLIST_IDENTIFIER: TodolistIdentifier = {
 
 /**
  * Main entry point for the Onestop to Basecamp Integration that contains the core logic for
- * adding/updating/deleting Basecamp Todos based on the rows on the Onestop
+ * adding/updating/deleting Basecamp Todos based on the rows on the Onestop as well as a function to
+ * create a menu item to give users the ability to manually trigger the import process.
```
andrewc5716, line 14: just a leftover comment here

andrewc5716 (approved):

PR #14, opened 2024-10-11
Dev/core logic update existing todos

jkzang:
### Update Existing Todos When OneStop Is Changed

This PR gives our code the ability to update existing todos whenever a change is made in the OneStop and the script is triggered. 

I manually tested the following scenarios and the updated code was able to successfully handle them by editing and moditying the todos in basecamp as expected:

1. Changing the location
2. Changing the date
3. Rearranging roles
4. Adding/removing new helper groups (i.e. food: )
5. Running the script after changes and again without changes ensuring that it doesn't make duplicate todos.

I also modified the document property methods to store the todo ids in a more descriptive way, separating them into different roles in an object.

### DIFF
--- a/src/main/basecamp.ts
+++ b/src/main/basecamp.ts
@@ -68,6 +68,18 @@ export function sendBasecampPutRequest(requestUrl: string, requestPayload: JsonO
     return JSON.parse(response.getContentText());
 }
 
+export function sendBasecampDeleteRequest(requestUrl: string): void {
+    const response: HTTPResponse = UrlFetchApp.fetch(requestUrl, {
+        method: HTTP_PUT_METHOD,
+        headers: getHeaders()
+    });
+
+    // Check if the response is ok (status code in the range 200-299)
+    if (response.getResponseCode() != 204) {
+        throw new Error(`HTTP error! Status: ${response.getResponseCode()}`);
+    }
+}
+
 /**
  * Performs a GET request for a Basecamp API, with built in pagination if applicable
  * 
--- a/src/main/basecamp.ts
+++ b/src/main/basecamp.ts
@@ -65,7 +65,14 @@ export function sendBasecampPutRequest(requestUrl: string, requestPayload: JsonO
         headers: getHeaders(),
         payload: JSON.stringify(requestPayload)
     });
-    return JSON.parse(response.getContentText());
+
+    const contentText = response.getContentText();
+
+    if (contentText) {
+        return JSON.parse(contentText);
+    } else {
+        return {};
+    }
 }
 
 /**
--- a/src/main/basecamp.ts
+++ b/src/main/basecamp.ts
@@ -66,7 +66,7 @@ export function sendBasecampPutRequest(requestUrl: string, requestPayload: JsonO
         payload: JSON.stringify(requestPayload)
     });
 
-    const contentText = response.getContentText();
+    const contentText: string = response.getContentText();
 
     if (contentText) {
         return JSON.parse(contentText);
--- a/src/main/basecamp.ts
+++ b/src/main/basecamp.ts
@@ -68,18 +68,6 @@ export function sendBasecampPutRequest(requestUrl: string, requestPayload: JsonO
     return JSON.parse(response.getContentText());
 }
 
-export function sendBasecampDeleteRequest(requestUrl: string): void {
-    const response: HTTPResponse = UrlFetchApp.fetch(requestUrl, {
-        method: HTTP_PUT_METHOD,
-        headers: getHeaders()
-    });
-
-    // Check if the response is ok (status code in the range 200-299)
-    if (response.getResponseCode() != 204) {
-        throw new Error(`HTTP error! Status: ${response.getResponseCode()}`);
-    }
-}
-
 /**
  * Performs a GET request for a Basecamp API, with built in pagination if applicable
  * 
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -4,7 +4,7 @@ import { RowBasecampMappingMissingError } from "./error/rowBasecampMappingMissin
 import { TodoIsMissingError } from "./error/todoIdMissingError";
 import { generateIdForRow, getBasecampTodoRequestsForRow, getId, getRowBasecampMapping, hasChanged, hasId, saveRow } from "./row";
 import { getEventRowsFromSpreadsheet } from "./scan";
-import { createTodo, TODOLIST_ID, updateTodo } from "./todos";
+import { createTodo, deleteTodo, TODOLIST_ID, updateTodo } from "./todos";
 
 const DEFAULT_TODOLIST_IDENTIFIER: TodolistIdentifier = {
     projectId: PROJECT_ID,
@@ -23,10 +23,10 @@ export function importOnestopToBasecamp(): void {
 
     for(const eventRow of eventRows) {
         if(hasId(eventRow)) {
-            console.log(`Row for ${eventRow.what.value} on ${eventRow.startTime} already exists! Processing as exiting row...\n`);
+            console.log(`Row for ${eventRow.what.value} on ${eventRow.startTime} already exists! Processing it as an existing row...\n`);
             processExistingRow(eventRow);
         } else {
-            console.log(`Row for ${eventRow.what.value} on ${eventRow.startTime} is new! Processing as a new row...\n`)
+            console.log(`Row for ${eventRow.what.value} on ${eventRow.startTime} is new! Processing it as a new row...\n`)
             processNewRow(eventRow);
         }
 
@@ -52,23 +52,30 @@ function processExistingRow(row: Row): void {
     if(hasChanged(row)) {
 
         Logger.log("Changes detected in row! Updateding associated todos to reflect changes...\n")
-        const basecampTodoRequests: Map<string, BasecampTodoRequest> = getBasecampTodoRequestsForRow(row);
-
+        
         const savedRowBasecampMapping: RowBasecampMapping | null = getRowBasecampMapping(row);
+        
+        if(savedRowBasecampMapping == null) {
+            throw new RowBasecampMappingMissingError("The rowBasecampMapping object is null! Unable to proceed with updating the todo!");
+        }
+        
+        const basecampTodoRequests: Map<string, BasecampTodoRequest> = getBasecampTodoRequestsForRow(row);
+        const currentRoles: string[] = Array.from(basecampTodoRequests.keys());
 
-        const roles: string[] = Array.from(basecampTodoRequests.keys());
+        const roleTodoIdMap = savedRowBasecampMapping.roleTodoIdMap
+        const originalRoles: string[] = Object.keys(roleTodoIdMap)
 
-        for (const role of roles) {
+        for (const role of currentRoles) {
 
-            let todoId = savedRowBasecampMapping?.roleTodoIdMap[role];
+            let todoId = roleTodoIdMap[role];
             let request = basecampTodoRequests.get(role);
 
             if(request == undefined) {
                 throw new BasecampRequestMissingError("Missing basecamp request!");
                 
             } else if (todoId == undefined) {
-                Logger.log(`Todo Id is undefined in the roleRodoIdMap! Creating new todo for ${row.what.value}...\n`)
-                createTodo(request, DEFAULT_TODOLIST_IDENTIFIER)
+                let newTodoId = createTodo(request, DEFAULT_TODOLIST_IDENTIFIER)
+                roleTodoIdMap[role] = newTodoId // Modify the roleTodoIdMap to reflect changes in the document properties
                 
             } else {
                 
@@ -77,17 +84,41 @@ function processExistingRow(row: Row): void {
                     todoId: todoId
                 }
                 
-                Logger.log(`Updating todo for ${row.what.value} (${row.startTime.getDate()})...\n`)
+                Logger.log(`Updating todo for ${row.what.value} (${row.startTime})...\n`)
                 updateTodo(request, todoIdentifier)
             }
         }   
 
-        const newRoles = Array.from(basecampTodoRequests.keys());
-        const oldRoles: string[] = roles.filter(role => !newRoles.includes(role));
+        deleteObsoleteTodos(originalRoles, currentRoles, roleTodoIdMap);
 
-        for(const oldRole of oldRoles) {
-            // Delete the todo associated with the old helper role
+        saveRow(row, roleTodoIdMap);
+    }
+}
+
+/**
+ * Deletes todos associated with helper groups that are no longer present on the event row.
+ * The first line checks whether there are any rows present in the original roles that are not present in the current roles (indicating an obsolete role)
+ * 
+ * @param originalRoles a list of all the original roles associated with the row including the lead role 
+ * @param currentRoles a list of all the current roles associated with the row including the lead role. This may be identical to the original roles
+ */
+function deleteObsoleteTodos(originalRoles: string[], currentRoles: string[], roleTodoIdMap: { [key: string]: string }): void {
+    Logger.log("Checking for obsolete roles...\n")
+    const obsoleteRoles: string[] = originalRoles.filter(role => !currentRoles.includes(role));
+    
+    for(const role of obsoleteRoles) {
+        Logger.log(`Found obsolete role: ${role}!\n`)
+        let todoId = roleTodoIdMap[role];
+
+        let todoIdentifier: TodoIdentifier = {
+            projectId: PROJECT_ID,
+            todoId: todoId
         }
+
+        Logger.log("Deleting todo...\n")
+        deleteTodo(todoIdentifier);
+
+        delete roleTodoIdMap[role]; // Modify the roleTodoIdMap to reflect changes in the document properties
     }
 }
 
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -47,12 +47,12 @@ function processExistingRow(row: Row): void {
     if(hasChanged(row)) {
 
         const currentRoleRequestMap: RoleRequestMap = getBasecampTodoRequestsForRow(row);
-        const currentRoleTodoIdMap: RoleTodoIdMap  = getRoleTodoIdMap(row);
+        const lastSavedRoleTodoIdMap: RoleTodoIdMap  = getRoleTodoIdMap(row);
 
-        deleteObsoleteTodos(currentRoleRequestMap, currentRoleTodoIdMap);
+        deleteObsoleteTodos(currentRoleRequestMap, lastSavedRoleTodoIdMap);
 
-        const newRoleTodoIdMap: RoleTodoIdMap = createTodosForNewRoles(currentRoleRequestMap, currentRoleTodoIdMap);
-        const existingRoleTodoIdMap: RoleTodoIdMap = updateTodosForExistingRoles(currentRoleRequestMap, currentRoleTodoIdMap)
+        const newRoleTodoIdMap: RoleTodoIdMap = createTodosForNewRoles(currentRoleRequestMap, lastSavedRoleTodoIdMap);
+        const existingRoleTodoIdMap: RoleTodoIdMap = updateTodosForExistingRoles(currentRoleRequestMap, lastSavedRoleTodoIdMap);
 
         const updatedRoleTodoIdMap: RoleTodoIdMap = {...existingRoleTodoIdMap, ...newRoleTodoIdMap};
 
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -54,9 +54,10 @@ function processExistingRow(row: Row): void {
         
         updateTodos(basecampTodoRequests, roleTodoIdMap)
 
-        deleteObsoleteTodos(basecampTodoRequests, roleTodoIdMap);
+        const obsoleteRoles = deleteObsoleteTodos(basecampTodoRequests, roleTodoIdMap);
+        const newRoleIdTodoIdMap = deleteObsoleteRolesFromRoleTodoIdMap(obsoleteRoles, roleTodoIdMap)
 
-        saveRow(row, roleTodoIdMap);
+        saveRow(row, newRoleIdTodoIdMap);
     }
 }
 
@@ -118,8 +119,9 @@ function getRoleTodoIdMap(row: Row) {
  * 
  * @param basecampTodoRequests a map associating role titles with BasecampTodoRequest objects
  * @param roleTodoIdMap a map associating an event's roles with todo ids
+ * @returns a string array of obsolete roles that were deleted
  */
-function deleteObsoleteTodos(basecampTodoRequests: Map<string, BasecampTodoRequest>, roleTodoIdMap: { [key: string]: string }): void {
+function deleteObsoleteTodos(basecampTodoRequests: Map<string, BasecampTodoRequest>, roleTodoIdMap: { [key: string]: string }): string[] {
 
     Logger.log("Checking for obsolete roles...\n")
 
@@ -137,9 +139,25 @@ function deleteObsoleteTodos(basecampTodoRequests: Map<string, BasecampTodoReque
         }
 
         deleteTodo(todoIdentifier);
+    }
 
+    return obsoleteRoles;
+}
+
+/**
+ * Modifies the roleTodoIdMap by removing obsolete roles.
+ * 
+ * @param obsoleteRoles a string array of obsolete roles no longer needed for an event
+ * @param roleTodoIdMap a map associating an event's roles with todo ids
+ * @returns the updated roleTodoIdMap without the obsolete roles
+ */
+function deleteObsoleteRolesFromRoleTodoIdMap(obsoleteRoles: string[], roleTodoIdMap: { [key: string]: string }):  { [key: string]: string } {
+
+    for(const role of obsoleteRoles) {
         delete roleTodoIdMap[role]; // Modify the roleTodoIdMap to reflect changes in the document properties
     }
+
+    return roleTodoIdMap
 }
 
 /**
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -1,14 +1,7 @@
-import { PROJECT_ID } from "./basecamp";
-import { RowBasecampMappingMissingError } from "./error/rowBasecampMappingMissingError";
-import { getRoleTodoIdMap } from "./role";
-import { generateIdForRow, getBasecampTodoRequestsForRow, getId, getRowBasecampMapping, hasChanged, hasId, saveRow } from "./row";
+import { getRoleTodoIdMap } from "./row";
+import { generateIdForRow, getBasecampTodoRequestsForRow, getId, hasChanged, hasId, saveRow } from "./row";
 import { getEventRowsFromSpreadsheet } from "./scan";
-import { createNewTodos, createTodo, createTodosForNewRoles, deleteObsoleteTodos, deleteTodo, deleteTodos, TODOLIST_ID, updateTodo, updateTodosForSurvivingRoles } from "./todos";
-
-export const DEFAULT_TODOLIST_IDENTIFIER: TodolistIdentifier = {
-    projectId: PROJECT_ID,
-    todolistId: TODOLIST_ID
-};
+import { createNewTodos, createTodosForNewRoles, deleteObsoleteTodos, updateTodosForExistingRoles } from "./todos";
 
 /**
  * Main entry point for the Onestop to Basecamp Integration that contains the core logic for
@@ -59,9 +52,9 @@ function processExistingRow(row: Row): void {
         deleteObsoleteTodos(currentRoleRequestMap, currentRoleTodoIdMap);
 
         const newRoleTodoIdMap: RoleTodoIdMap = createTodosForNewRoles(currentRoleRequestMap, currentRoleTodoIdMap);
-        const survivingRoleTodoIdMap: RoleTodoIdMap = updateTodosForSurvivingRoles(currentRoleRequestMap, currentRoleTodoIdMap)
+        const existingRoleTodoIdMap: RoleTodoIdMap = updateTodosForExistingRoles(currentRoleRequestMap, currentRoleTodoIdMap)
 
-        const updatedRoleTodoIdMap: RoleTodoIdMap = {...survivingRoleTodoIdMap, ...newRoleTodoIdMap};
+        const updatedRoleTodoIdMap: RoleTodoIdMap = {...existingRoleTodoIdMap, ...newRoleTodoIdMap};
 
         saveRow(row, updatedRoleTodoIdMap);
     }
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -49,52 +49,120 @@ function processExistingRow(row: Row): void {
 
     if(hasChanged(row)) {
 
-        const roleTodoIdMap = getRoleTodoIdMap(row);
         const basecampTodoRequests: Map<string, BasecampTodoRequest> = getBasecampTodoRequestsForRow(row);
+        const existingRoleTodoIdMap: { [key: string]: string }  = getRoleTodoIdMap(row);
         
-        updateTodos(basecampTodoRequests, roleTodoIdMap)
+        const currentRoles: string[] = getCurrentEventRoles(basecampTodoRequests);
+        const originalRoles: string[] = getOriginalEventRoles(existingRoleTodoIdMap);
 
-        const obsoleteRoles = deleteObsoleteTodos(basecampTodoRequests, roleTodoIdMap);
-        const newRoleIdTodoIdMap = deleteObsoleteRolesFromRoleTodoIdMap(obsoleteRoles, roleTodoIdMap)
+        const newRoles: string[] = getNewRoles(currentRoles, originalRoles);
+        const newRoleTodoIdMap: { [key: string]: string } = createTodosForNewRoles(newRoles, basecampTodoRequests);
+
+        const existingRoles: string[] = getExistingRoles(currentRoles, originalRoles);
+        updateTodosForExistingRoles(existingRoles, basecampTodoRequests, existingRoleTodoIdMap)
+
+        const obsoleteRoles: string[] = getObsoleteRoles(currentRoles, originalRoles);
+        deleteObsoleteTodos(obsoleteRoles, existingRoleTodoIdMap);
+
+        const newRoleIdTodoIdMap = updateRoleTodoIdMap(obsoleteRoles, newRoleTodoIdMap, existingRoleTodoIdMap)
 
         saveRow(row, newRoleIdTodoIdMap);
     }
 }
 
 /**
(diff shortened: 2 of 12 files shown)


src/main/basecamp.ts
```diff
@@ -68,6 +68,18 @@ export function sendBasecampPutRequest(requestUrl: string, requestPayload: JsonO
     return JSON.parse(response.getContentText());
 }
 
+export function sendBasecampDeleteRequest(requestUrl: string): void {
```
andrewc5716, line 71: technically this isn't a basecamp delete then, this is still using a basecamp put. the wrapper function outside of this should then be appropriately named as delete todo (which it is)

src/main/main.ts
```diff
@@ -34,8 +39,87 @@ export function importOnestopToBasecamp(): void {
     deleteOldRows(processedRowIds);
 }
 
+/**
+ * Processes existing event rows by checking whether the row has been modified and if so indiscriminately
+ * updates all the tasks with the new information
+ * 
+ * @param row 
+ */
 function processExistingRow(row: Row): void {
```
andrewc5716, line 48: this is a huge function 🥲 any way to break this up into smaller functions?

src/main/row.ts
```diff
@@ -210,44 +211,34 @@ function toHexString(byteArray: number[]): string {
...
  * 
  * @param row row to construct the BasecampTodoRequest for
- * @returns BasecampTodoRequest
+ * @param basecampTodoRequests the Map that associates the roleTitle with the request
  */
-export function getBasecampTodoForLeads(row: Row): BasecampTodoRequest | undefined {
-    const basecampTodoContent: string = `Lead: ${row.what.value}`;
+export function getBasecampTodoForLeads(row: Row, basecampTodoRequests: Map<string, BasecampTodoRequest>): void {
+
+    const roleTitle = "Lead"
```
andrewc5716, line 231: let's make this a constant defined at the top of this class

src/main/basecamp.ts
andrewc5716: so basically i think it's ok to not have this function, and delete todo should just directly use basecamp delete

src/main/main.ts
```diff
@@ -34,8 +39,87 @@ export function importOnestopToBasecamp(): void {
...
+
+        let todoIdentifier: TodoIdentifier = {
+            projectId: PROJECT_ID,
+            todoId: todoId
+        }
+
+        Logger.log("Deleting todo...\n")
+        deleteTodo(todoIdentifier);
+
+        delete roleTodoIdMap[role]; // Modify the roleTodoIdMap to reflect changes in the document properties
```
andrewc5716, line 121: as much as possible, we should avoid hidden side-effects with functions (here the roleTodoIdMap is modified)

src/main/main.ts
```diff
@@ -34,8 +39,87 @@ export function importOnestopToBasecamp(): void {
...
+}
+
+/**
+ * Deletes todos associated with helper groups that are no longer present on the event row.
+ * The first line checks whether there are any rows present in the original roles that are not present in the current roles (indicating an obsolete role)
+ * 
+ * @param originalRoles a list of all the original roles associated with the row including the lead role 
+ * @param currentRoles a list of all the current roles associated with the row including the lead role. This may be identical to the original roles
+ */
+function deleteObsoleteTodos(originalRoles: string[], currentRoles: string[], roleTodoIdMap: { [key: string]: string }): void {
```
andrewc5716, line 105: this function can be broken into different parts:
1. getObsoleteRoles: takes original and current roles, returns list of obsolete roles
2. getObsoleteTodos: takes list of obsolete roles and roleTodoIdMap, returns todoIds
3. deleteObsoleteTodos: takes todoIds, deletes them
4. removeDeletedRoles: takes roleTodoIdMap, list of obsolete roles, returns a copy of the map with the roles deleted

src/main/main.ts
```diff
@@ -34,8 +39,87 @@ export function importOnestopToBasecamp(): void {
...
+            throw new RowBasecampMappingMissingError("The rowBasecampMapping object is null! Unable to proceed with updating the todo!");
+        }
+        
+        const basecampTodoRequests: Map<string, BasecampTodoRequest> = getBasecampTodoRequestsForRow(row);
+        const currentRoles: string[] = Array.from(basecampTodoRequests.keys());
+
+        const roleTodoIdMap = savedRowBasecampMapping.roleTodoIdMap
+        const originalRoles: string[] = Object.keys(roleTodoIdMap)
+
+        for (const role of currentRoles) {
```
andrewc5716, line 68: this could be broken up as well:
1. getNewRoles 
2. getExistingRoles
3. createTodosForNewRoles - needs to return todoIds to avoid side effects
4. updateTodosForExistingRoles
5. addNewRoles (to roleTodoIdMap), returns a copy with the new roles added

src/main/basecamp.ts
jkzang: Got it

src/main/main.ts
jkzang: Will do

jkzang: It's broken up now

src/main/basecamp.ts
jkzang: It's deleted

src/main/row.ts
jkzang: It's done

src/main/main.ts
jkzang: K the PR is updated with this as well

src/main/main.ts
jkzang: Sounds good, the new code reflects this change request as well

src/main/main.ts
```diff
@@ -20,8 +23,10 @@ export function importOnestopToBasecamp(): void {
 
     for(const eventRow of eventRows) {
         if(hasId(eventRow)) {
+            console.log(`Row for ${eventRow.what.value} on ${eventRow.startTime} already exists! Processing it as an existing row...\n`);
             processExistingRow(eventRow);
         } else {
+            console.log(`Row for ${eventRow.what.value} on ${eventRow.startTime} is new! Processing it as a new row...\n`)
```
jbwong05, line 29: For logging, we need to use `Logger.log()` for Google Apps Scripts

src/main/scan.ts
```diff
@@ -36,11 +36,14 @@ interface CellData {
...
 export function getEventRowsFromSpreadsheet(): Row[] {
+    console.log("Getting event rows from spreadsheet...\n");
+    
     const tabs: Sheet[] = getAllSpreadsheetTabs();
     const dailyActiveTabs: Sheet[] = getActiveDailyTabs(tabs);
     // Fetches all event rows from all of the daily active tabs
     const eventRows: Row[] = dailyActiveTabs.flatMap((dailyActiveTab) => getRowsWithEvents(dailyActiveTab));
-
+    
+    console.log(`Found ${eventRows.length} active rows from the spreadsheet...`)
```
jbwong05, line 46: `Logger.log()` should be used here too

src/main/types.d.ts
```diff
@@ -32,7 +32,7 @@ declare interface Row {
 
 declare interface RowBasecampMapping {
   rowHash: string,
-  basecampTodoIds: string[]
+  roleTodoIdMap: { [key: string]: string }
```
jbwong05, line 35: Maybe it's worth maybe a new type for this so we don't have `{ [key: string]: string }` floating around everywhere

src/main/propertiesService.ts
```diff
@@ -112,3 +113,23 @@ export function deleteAllDocumentProperties(): void {
...
+ * @function logDocumentProperties
+ * @returns {void}
+ *
+ * @example
+ * logDocumentProperties();
+ */
+export function logDocumentProperties() {
+    var allProperties = documentProperties.getProperties();
+    for (var key in allProperties) {
+      Logger.log(key + ': ' + allProperties[key]);
```
jbwong05, line 132: Nit: I know this is just a function for debugging but this code is missing type definitions

jbwong05:
Left some brief comments after just a brief glance over but haven't had time to complete a full in depth review especially over any of the new logic because it's kind of late right now. Will circle back on completing a more in depth review later

src/main/main.ts
jkzang: createTodosForNewRoles - needs to return todoIds to avoid side effects

I can still implement this, but not sure why returning todoids avoids side effects?

jkzang: Oh wait, actually I know why now

jkzang: Okay I implemented all the new functions

src/main/main.ts
jkzang: Actually, I think console.log still works, but yeah I agree, for consistency, we should just use Logger.log

jkzang: I updated all the instances of console.log to use Logger

src/main/types.d.ts
jkzang: Yeah, I was getting bothered by that as well. Thanks for bringing it up. I created a simple type alias in the types file.

src/main/propertiesService.ts
jkzang: Updated logDocumentProperties() with type definitions

jkzang:
Made a lot of changes and also reorganized the code so that all the helper functions are not cluttering the main.ts file. Please take a look when you have the chance.

src/main/basecamp.ts
```diff
@@ -65,7 +65,14 @@ export function sendBasecampPutRequest(requestUrl: string, requestPayload: JsonO
         headers: getHeaders(),
         payload: JSON.stringify(requestPayload)
     });
-    return JSON.parse(response.getContentText());
+
+    const contentText = response.getContentText();
```
jbwong05, line 69: Type definition is missing here

src/main/main.ts
```diff
@@ -1,9 +1,11 @@
 import { PROJECT_ID } from "./basecamp";
-import { generateIdForRow, getBasecampTodoRequestsForRow, getId, hasId, saveRow } from "./row";
+import { RowBasecampMappingMissingError } from "./error/rowBasecampMappingMissingError";
+import { getRoleTodoIdMap } from "./role";
+import { generateIdForRow, getBasecampTodoRequestsForRow, getId, getRowBasecampMapping, hasChanged, hasId, saveRow } from "./row";
 import { getEventRowsFromSpreadsheet } from "./scan";
-import { createTodo, TODOLIST_ID } from "./todos";
+import { createNewTodos, createTodo, createTodosForNewRoles, deleteObsoleteTodos, deleteTodo, deleteTodos, TODOLIST_ID, updateTodo, updateTodosForSurvivingRoles } from "./todos";
```
jbwong05, line 6: There are some unused imports here after the refactoring

src/main/role.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @param row a list of all the current roles associated with the row including the lead role. This may be identical to the original roles
+ * @returns a map that associates role titles with basecamp todo ids
+ */
+export function getRoleTodoIdMap(row: Row) {
+    const savedRowBasecampMapping: RowBasecampMapping | null = getRowBasecampMapping(row);
+    if(savedRowBasecampMapping == null) {
+        throw new RowBasecampMappingMissingError("The rowBasecampMapping object is null! Unable to proceed with updating the todo!");
+    }
+    return savedRowBasecampMapping.roleTodoIdMap
+}
```
jbwong05, line 23: This function takes a Row as a parameter. Should it belong in the `row.ts` file?

src/main/role.ts
```diff
@@ -0,0 +1,64 @@
...
+ * 
+ * @param row a list of all the current roles associated with the row including the lead role. This may be identical to the original roles
+ * @returns a map that associates role titles with basecamp todo ids
+ */
+export function getRoleTodoIdMap(row: Row) {
+    const savedRowBasecampMapping: RowBasecampMapping | null = getRowBasecampMapping(row);
+    if(savedRowBasecampMapping == null) {
+        throw new RowBasecampMappingMissingError("The rowBasecampMapping object is null! Unable to proceed with updating the todo!");
+    }
+    return savedRowBasecampMapping.roleTodoIdMap
```
jbwong05, line 22: Missing semicolon

src/main/role.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Gets the roleTodoIdMap object from the RowBasecampMapping object.
+ * Used for downstream processing
+ * 
+ * @param row a list of all the current roles associated with the row including the lead role. This may be identical to the original roles
+ * @returns a map that associates role titles with basecamp todo ids
+ */
+export function getRoleTodoIdMap(row: Row) {
+    const savedRowBasecampMapping: RowBasecampMapping | null = getRowBasecampMapping(row);
+    if(savedRowBasecampMapping == null) {
```
jbwong05, line 19: Do we need to use triple equal sign (===) here? I'm not super familiar with all the nuances and differences between the two

src/main/types.d.ts
```diff
@@ -30,9 +30,11 @@ declare interface Row {
   readonly notes: Text
 }
 
+type RoleTodoIdMap = { [key: string]: string };
```
jbwong05, line 33: We should probably add a comment here to document what the key and value are for this map

src/main/todos.ts
```diff
@@ -1,22 +1,38 @@
...
 const TODO_JSON_PATH: string = '/todos' + JSON_PATH;
+const TRASHED_TODO_JSON_PATH: string = '/status/trashed' + JSON_PATH;
+
 export const TODOLIST_ID: string= "7865336721";
 
+const DEFAULT_TODOLIST_IDENTIFIER: TodolistIdentifier = {
+    projectId: PROJECT_ID,
+    todolistId: TODOLIST_ID
+};
+
```
jbwong05, line 18: It looks like this is a duplicate of the definition that is being exported from `main.ts`

src/main/role.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * Gets a list of roles from a roleRequestMap map
+
+ * @param roleRequestMap a map associating role titles with BasecampTodoRequest objects
+ */
+export function getCurrentEventRoles(currentRoleRequestMap: RoleRequestMap): string[] {
+    return Object.keys(currentRoleRequestMap);
+}
```
jbwong05, line 41: If these functions are just helper functions for the `getRoles` function in this file, they do not need to be exported

src/main/row.ts
```diff
@@ -367,11 +366,11 @@ function getHelperGroups(row: Row): HelperGroup[] {
             const [role, helperNameList] = helperLine.split(COLON_DELIM);
             const trimmedHelperNameList: string = helperNameList.trim();
             helperGroups.push(getHelperGroupFromNameList(trimmedHelperNameList, role));
-        } else {
+            
+        } else if(helperLine != "") {
```
jbwong05, line 370:
```suggestion
        } else if(helperLine !== "") {
```

src/main/role.ts
jbwong05, line 19:
```suggestion
    if(savedRowBasecampMapping === null) {
```

src/main/role.ts
jbwong05, line 41:
```suggestion
function getOriginalEventRoles(roleTodoIdMap: RoleTodoIdMap): string[] {
    return Object.keys(roleTodoIdMap);
}

/**
 * Gets a list of roles from a roleRequestMap map

 * @param roleRequestMap a map associating role titles with BasecampTodoRequest objects
 */
 function getCurrentEventRoles(currentRoleRequestMap: RoleRequestMap): string[] {
    return Object.keys(currentRoleRequestMap);
}
```
If these functions are just helper functions for the getRoles function in this file, they do not need to be exported

src/main/basecamp.ts
jbwong05, line 69:
```suggestion
    const contentText:string = response.getContentText();
```

src/main/role.ts
```diff
@@ -0,0 +1,64 @@
+import { RowBasecampMappingMissingError } from "./error/rowBasecampMappingMissingError";
+import { getRowBasecampMapping } from "./row";
+
+export enum RoleStatus {
+    OBSOLETE,
+    SURVIVING,
+    NEW
```
jbwong05, line 7: I was a little confused at first about what it meant for a role to be obsolete or surviving. Would something like `Removed` and `Unchanged or Existing` be more descriptive?

src/main/todos.ts
```diff
@@ -1,22 +1,38 @@
...
+export function createTodo(request: BasecampTodoRequest | undefined, todolistIdentifier: TodolistIdentifier): string {
+    
+    if(request == undefined) {
+        throw new BasecampRequestMissingError("Missing basecamp request!");
+    } else {
+        Logger.log(`Creating new todo: "${request.content}"...\n`)
+        const rawTodoResponse: JsonData = sendBasecampPostRequest(getCreateTodoUrl(todolistIdentifier), request);
+        const todoResponse: BasecampTodoResponse = rawTodoResponse as BasecampTodoResponse;
+        return todoResponse.id;
+    }
```
jbwong05, line 35: When would a request be undefined? If for some reason it does have a possibility of being undefined, I feel like we should be checking for that before even calling this function to send the POST request

src/main/todos.ts
```diff
@@ -26,8 +42,26 @@ export function createTodo(todo: BasecampTodoRequest, todolistIdentifier: Todoli
...
-    sendBasecampPutRequest(getUpdateTodoUrl(todoIdentifier), todo);
+export function updateTodo(request: BasecampTodoRequest | undefined, todoIdentifier: TodoIdentifier): void {
+
+    if(request == undefined) {
+        throw new BasecampRequestMissingError("Missing basecamp request!");
+    } else {
+        Logger.log(`Updating existing todo: "${request.content}"...\n`)
+        sendBasecampPutRequest(getUpdateTodoUrl(todoIdentifier), request);
+    }
+}
```
jbwong05, line 53: Same comment as above for the `createTodo` function with the request being undefined - https://github.com/andrewc5716/onestop-basecamp-integration/pull/14/files#r1800223947

src/main/todos.ts
```diff
@@ -59,4 +97,141 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+/**
+ * Gets a list of obsolete todo ids based off of the obsolete roels roles by comparing the original and current roles together
+ * 
+ * @param obsoleteRoles an array of obsolete roles
+ * @param roleTodoIdMap a map associating an event's roles with todo ids
+ * @returns an array of obsolete roles
+ */
+export function getObsoleteTodosIds(obsoleteRoles: string[], roleTodoIdMap: RoleTodoIdMap): string[] {
+
+    const obsoleteTodoIds = []
```
jbwong05, line 170:
```suggestion
    const obsoleteTodoIds = [];
```

src/main/todos.ts
```diff
@@ -59,4 +97,141 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+    
+    for(const role of obsoleteRoles) {
+        Logger.log(`Found obsolete role: ${role}!\n`)
```
jbwong05, line 173:
```suggestion
        Logger.log(`Found obsolete role: ${role}!\n`);
```

src/main/todos.ts
```diff
@@ -59,4 +97,141 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+        let todoId = roleTodoIdMap[role];
+        obsoleteTodoIds.push(todoId);
+    }
+
+    return obsoleteTodoIds
```
jbwong05, line 178:
```suggestion
    return obsoleteTodoIds;
```

src/main/todos.ts
```diff
@@ -59,4 +97,141 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+    const survivingRoles: string[] = getRoles(roleRequestMap, roleTodoIdMap, RoleStatus.SURVIVING);
+
+    for(const role of survivingRoles) {
+        let survivingTodoId = roleTodoIdMap[role];
+        let request = roleRequestMap[role];
+
+        let todoIdentifier: TodoIdentifier = {
+            projectId: PROJECT_ID,
+            todoId: survivingTodoId
+        }
```
jbwong05, line 201:
```suggestion
        };
```

src/main/todos.ts
```diff
@@ -59,4 +97,141 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+ * @returns an object mapping new roles with their newly created todo ids
+ */
+export function createTodosForNewRoles(roleRequestMap: RoleRequestMap, roleTodoIdMap: RoleTodoIdMap): RoleTodoIdMap {
+    Logger.log("Checking for new roles...\n");
+
+    const newRoles: string[] = getRoles(roleRequestMap, roleTodoIdMap, RoleStatus.NEW);
+    const newRoleTodoIdMap: RoleTodoIdMap = {};
+    
+    if(newRoles.length > 0) {
+        Logger.log(`New role(s) detected: ${newRoles}\n`)
```
jbwong05, line 225:
```suggestion
        Logger.log(`New role(s) detected: ${newRoles}\n`);
```

src/main/todos.ts
```diff
@@ -59,4 +97,141 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+        for(const role of newRoles) {
+            let request = roleRequestMap[role];
+            let newTodoId = createTodo(request, DEFAULT_TODOLIST_IDENTIFIER);
+            newRoleTodoIdMap[role] = newTodoId;
+        }
+
+    } else {
+        Logger.log("No new roles detected!\n")
```
jbwong05, line 233:
```suggestion
        Logger.log("No new roles detected!\n");
```

src/main/main.ts
jkzang: Removed unused imports

src/main/role.ts
jkzang: Moved getRoleTodoIdMap() to the row.ts file

src/main/role.ts
jkzang: I updated the enum to use EXISTING instead of SURVIVING and REMOVED instead of OBSOLETE

src/main/todos.ts
jkzang: I did run into the request being undefined at some points, but that was only because earlier in the function the code was trying to get the value for an unexpected role, but that was due to a bug that I fixed even more upstream in the logic in previous commits so we shouldn't need to worry about that. I agree, it makes more sense to check for undefined requests if we are going to check for them before running the createTodo function.

jkzang:
@jbwong05 Thanks for the thorough code review! I addressed each one of your points and made the recommended changes

jkzang:
@andrewc5716 do you think that this is ready to be merged?

src/main/row.ts
```diff
@@ -210,45 +214,40 @@ function toHexString(byteArray: number[]): string {
...
-    }
-
-    const helpersBasecampTodoRequest: BasecampTodoRequest[] = getBasecampTodosForHelpers(row);
+    const leadsRoleRequestMap: RoleRequestMap = getBasecampTodoForLeads(row);
+    const helperRoleRequestMap: RoleRequestMap = getBasecampTodosForHelpers(row);
 
-    if(helpersBasecampTodoRequest.length > 0) {
-        basecampTodoRequests = basecampTodoRequests.concat(helpersBasecampTodoRequest);
-    }
+    const allRoleRequestMap = {...leadsRoleRequestMap, ...helperRoleRequestMap};
```
jbwong05, line 224:
```suggestion
    const allRoleRequestMap: RoleRequestMap = {...leadsRoleRequestMap, ...helperRoleRequestMap};
```
Must have missed from before

src/main/row.ts
```diff
@@ -338,14 +337,15 @@ export function getBasecampTodosForHelpers(row: Row): BasecampTodoRequest[] {
         const basecampDueDate: string = getBasecampDueDate(row);
 
         if(assigneeIds.length > 0) {
-            basecampTodoRequests.push(getBasecampTodoRequest(basecampTodoContent, basecampTodoDescription, 
-                assigneeIds, leadIds, false, basecampDueDate));
+            const basecampTodoRequest = getBasecampTodoRequest(basecampTodoContent, basecampTodoDescription, assigneeIds, leadIds, false, basecampDueDate);
```
jbwong05, line 340:
```suggestion
            const basecampTodoRequest: BasecampTodoRequest = getBasecampTodoRequest(basecampTodoContent, basecampTodoDescription, assigneeIds, leadIds, false, basecampDueDate);
```
Also must have missed from before

src/main/todos.ts
```diff
@@ -59,4 +88,153 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+ * Sends requests to create new Basecamp Todos
+ * 
+ * @param basecampRequests array of BasecampTodoRequest objects to send
+ * @returns a map that associates role titles with basecamp todo ids
+ */
+export function createNewTodos(roleRequestMap: RoleRequestMap): RoleTodoIdMap {
+    const roleTodoIdMap: RoleTodoIdMap = {};
+
+    Object.keys(roleRequestMap).forEach( role => {
+        let request = roleRequestMap[role];
```
jbwong05, line 103:
```suggestion
        let request: BasecampTodoRequest = roleRequestMap[role];
```
Sorry also must have missed from before

src/main/todos.ts
```diff
@@ -59,4 +88,153 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+}
+
+/**
+ * Gets a list of obsolete todo ids based off of the obsolete roels roles by comparing the original and current roles together
+ * 
+ * @param obsoleteRoles an array of obsolete roles
+ * @param roleTodoIdMap a map associating an event's roles with todo ids
+ * @returns an array of obsolete roles
+ */
+export function getObsoleteTodosIds(obsoleteRoles: string[], roleTodoIdMap: RoleTodoIdMap): string[] {
```
jbwong05, line 159: Because the role status of `obselete` was renamed, functions and variables like this should be renamed accordingly

src/main/todos.ts
```diff
@@ -59,4 +88,153 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+
+    const obsoleteTodoIds = [];
+    
+    for(const role of obsoleteRoles) {
+        Logger.log(`Found removed role: ${role}!\n`);
+        let todoId = roleTodoIdMap[role];
+        obsoleteTodoIds.push(todoId);
+    }
```
jbwong05, line 167:
```suggestion
    const obsoleteTodoIds: string[] = [];
    
    for(const role of obsoleteRoles) {
        Logger.log(`Found removed role: ${role}!\n`);
        let todoId: string = roleTodoIdMap[role];
        obsoleteTodoIds.push(todoId);
    }
```

src/main/todos.ts
```diff
@@ -59,4 +88,153 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+ * @returns an object mapping surviving roles with their existing todo ids
+ */
+export function updateTodosForExistingRoles(roleRequestMap: RoleRequestMap, roleTodoIdMap: RoleTodoIdMap): RoleTodoIdMap {
+    Logger.log("Updating todos for existing roles...");
+    const existingRoleTodoIdMap: RoleTodoIdMap = {};
+    const existingRoles: string[] = getRoles(roleRequestMap, roleTodoIdMap, RoleStatus.EXISTING);
+
+    for(const role of existingRoles) {
+        let existingTodoId = roleTodoIdMap[role];
+        let request = roleRequestMap[role];
```
jbwong05, line 187:
```suggestion
        let existingTodoId: string = roleTodoIdMap[role];
        let request: BasecampTodoRequest = roleRequestMap[role];
```

src/main/todos.ts
```diff
@@ -59,4 +88,153 @@ export function getBasecampTodoRequest(content: string, description: string, ass
...
+        Logger.log(`New role(s) detected: ${newRoles}\n`);
+        for(const role of newRoles) {
+            let request = roleRequestMap[role];
+
+            if(request === undefined) {
+                throw new BasecampRequestMissingError("Missing basecamp request!");
+            }
+
+            let newTodoId = createTodo(request, DEFAULT_TODOLIST_IDENTIFIER);
+            newRoleTodoIdMap[role] = newTodoId;
```
jbwong05, line 232:
```suggestion
            let request: BasecampTodoRequest = roleRequestMap[role];

            if(request === undefined) {
                throw new BasecampRequestMissingError("Missing basecamp request!");
            }

            let newTodoId: string = createTodo(request, DEFAULT_TODOLIST_IDENTIFIER);
            newRoleTodoIdMap[role] = newTodoId;
```

jkzang:
@jbwong05 I added the missing type defs from your latest round of review

jkzang:
@andrewc5716 I did some testing and I didn't see any issues with the edit features on my branch

src/main/todos.ts
jbwong05: I think it's ok. Thanks for the clarification

jbwong05:
@jkzang Thanks for thoroughly addressing all of my comments. Looks a lot better from my side (although I haven't had time to actually deploy and test this branch myself). Awaiting @andrewc5716's 👀 and approval since he referenced some bug on this branch after the refactoring in the chat this morning

src/main/role.ts
```diff
@@ -0,0 +1,46 @@
...
+ * 
+ * @param roleRequestMap a map associating an event's roles with api request bodies
+ * @param currentRoleTodoIdMap a map associating an event's roles to existing todo ids, that is currently saved in the document properties
+ * @returns an array of obsolete roles
+ */
+export function getRoles(roleRequestMap: RoleRequestMap, currentRoleTodoIdMap: RoleTodoIdMap, roleStatus: RoleStatus): string[] {
+    const currentRoles: string[] = getCurrentEventRoles(roleRequestMap);
+    const originalRoles: string[] = getOriginalEventRoles(currentRoleTodoIdMap);
+
+    switch(roleStatus) {
```
andrewc5716, line 36: best practice is to avoid "flag" arguments for methods. this is because it unnecessarily complicates a function (even though in this example it's not that complex), since the alternative is to have 3 different simple functions, appropriately named (getRemovedRoles, getExistingRoles, getNewRoles) which is far more clear.
and best practice is also to minimize the number of arguments in a function as much as possible, as it makes testing more work.

src/main/main.ts
```diff
@@ -34,8 +31,33 @@ export function importOnestopToBasecamp(): void {
...
+ * 
+ * @param row 
+ */
 function processExistingRow(row: Row): void {
-    // Will be implemented as part of https://3.basecamp.com/4474129/buckets/38736474/todos/7717428992
+
+    if(hasChanged(row)) {
+
+        const currentRoleRequestMap: RoleRequestMap = getBasecampTodoRequestsForRow(row);
+        const currentRoleTodoIdMap: RoleTodoIdMap  = getRoleTodoIdMap(row);
```
andrewc5716, line 50: small nitpick: i happen to know roleTodoIdMap is what's last saved, but since the name here has "current", if i didn't know it's what's saved, it could be confusing. what do you think of a name like "lastSaved"?

src/main/main.ts
andrewc5716: nice i like it, much cleaner

src/main/role.ts
jkzang: I see your point how avoiding flag arguments for methods and breaking out this function into 3 separate methods makes things clearer. However, the reason why I merged them in the first place is that I didn't like how I would have to repeat the getCurrentEventRoles() and getOriginalEventRoles() methods over and over again:

`/**
 * Gets a list of obsolete roles by comparing the original and current roles
 * 
 * @param roleRequestMap a map associating an event's roles with API request bodies
 * @param lastSavedRoleTodoIdMap a map associating an event's roles to existing todo IDs, that is currently saved in the document properties
 * @returns an array of removed roles
 */
export function getRemovedRoles(roleRequestMap: RoleRequestMap, lastSavedRoleTodoIdMap: RoleTodoIdMap): string[] {
    const currentRoles: string[] = getCurrentEventRoles(roleRequestMap);
    const originalRoles: string[] = getOriginalEventRoles(lastSavedRoleTodoIdMap);
    return originalRoles.filter(role => !currentRoles.includes(role));
}

/**
 * Gets a list of existing roles by comparing the original and current roles
 * 
 * @param roleRequestMap a map associating an event's roles with API request bodies
 * @param lastSavedRoleTodoIdMap a map associating an event's roles to existing todo IDs, that is currently saved in the document properties
 * @returns an array of existing roles
 */
export function getExistingRoles(roleRequestMap: RoleRequestMap, lastSavedRoleTodoIdMap: RoleTodoIdMap): string[] {
    const currentRoles: string[] = getCurrentEventRoles(roleRequestMap);
    const originalRoles: string[] = getOriginalEventRoles(lastSavedRoleTodoIdMap);
    return currentRoles.filter(role => originalRoles.includes(role));
}

/**
 * Gets a list of new roles by comparing the original and current roles
 * 
 * @param roleRequestMap a map associating an event's roles with API request bodies
 * @param lastSavedRoleTodoIdMap a map associating an event's roles to existing todo IDs, that is currently saved in the document properties
 * @returns an array of new roles
 */
export function getNewRoles(roleRequestMap: RoleRequestMap, lastSavedRoleTodoIdMap: RoleTodoIdMap): string[] {
    const currentRoles: string[] = getCurrentEventRoles(roleRequestMap);
    const originalRoles: string[] = getOriginalEventRoles(lastSavedRoleTodoIdMap);
    return currentRoles.filter(role => !originalRoles.includes(role));
}`

Is there a way to get around that?

jkzang: I can still break out the getRoles method into the three functions, but I'm just wondering if there's a way to address the repeated code? I could try calling these methods two levels up and pass the currentRoles and originalRoles arrays into the calling functions for getNewRoles, getExistingRoles, and getRemovedRoles however that would increase the argument counts for each of the calling functions to 4, which I think is a lot

jkzang: I broke out getRoles() into the three methods that you suggested

src/main/main.ts
jkzang: That makes sense, I changed the variable names in my latest commit

jbwong05 (approved):
Approving as it looks like all of Andrew's comments have been addressed. @jkzang Are you going to merge https://github.com/andrewc5716/onestop-basecamp-integration/pull/16 into this branch first before merging this PR into master?

jkzang:
Thanks! I think I'll merge them separately

src/main/role.ts
andrewc5716: you're right that there's repeated code. and that is a good principle too (Don't Repeat Yourself), but in this case it's only 2 lines of code repeated, which is pretty insignificant. my opinion here is that the clarity that comes from removing the flag argument outweighs the 2 lines of repeated code.
i also thought about the possibility of calling those functions earlier. you're also right that passing in 4 things is a lot. the general way to get around this is to call it up earlier and wrap all 4 things necessary into its own object, and pass that. in this case, it might be a bit too over-engineered (unless this type of object really is used a lot), and to solve for just 2 lines of repeated code, it's probably not necessary

PR #15, opened 2024-10-11
Add github actions deployment and move prod secrets to script properties

jbwong05:
## Description
This PR adds the github action definition to build and deploy our code to Google Apps Script. This also adds the ability to fetch the Basecamp secrets from script properties for the prod deployment

## Github actions
* The action should dynamically generate the .clasprc.json and the .clasp.json. All parameters required for generating these files need to be stored in the github repo secrets (these have not been added yet; ideally we would have a service account)
* Also added a github action to build the project on PRs ensuring the project builds successfully before it is merged
* There is limited ability to be able to test these deployment scripts locally so this will still need to be tested
  * Tested locally with [act](https://nektosact.com/introduction.html)
    * To test with act, install act and run `act push --secret-file github.secrets` for testing the deployment action or `act pull_request` to test the building action in the root dir of the project (act depends on docker)

`github.secrets file`:
```
ACCESS_TOKEN=
ID_TOKEN=
REFRESH_TOKEN=
CLIENT_ID=
CLIENT_SECRET=
SCRIPT_ID=
```

## Basecamp secrets from script properties
Getting different secret definitions to work depending on the environment while not leaking secrets was kinda tricky. webpack does not have an exact equivalent of `#ifdef` like in C so had to make due with some suggestions from ChatGPT. Basically we can specify a parameter while "building" to designate the environment (this means we now have separate build commands for dev/prod). Based on the parameter, webpack will dynamically import the dev or prod credentials. We cannot just import both or else both definitions would get copied into the final script potentially leaking credentials. The `/* webpackMode: "eager" */` annotation is required to ensure webpack keeps the dynamic imports within the same bundle (Google Apps Script did not like multiple bundles for some reason; this is ok since the dynamic fetching of the prod secrets depends on code from the main bundle anyways)

### DIFF
--- a/.github/workflows/build-deploy.yaml
+++ b/.github/workflows/build-deploy.yaml
@@ -1,0 +1,46 @@
+name: Build and Deploy
+
+on:
+    push:
+        branches:
+            master
+
+jobs:
+    build-deploy:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Set up Node.js
+              uses: actions/setup-node@v3
+              with:
+                node-version: '22.8.0'
+            - name: Install clasp
+              run: npm install -g @google/clasp
+            - name: Checkout code
+              uses: actions/checkout@v3
+            - name: Install project dependencies
+              run: npm install
+            - name: Generate .clasprc.json
+              run: |
+                echo '{
+                    "token": {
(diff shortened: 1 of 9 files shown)


andrewc5716 (approved):
this looks complicated.. nice getting this to work

PR #21, opened 2024-10-29
Add more row unit tests

jbwong05:
Small PR that adds some more Row unit tests for the following functions:
* getMetadata()
* generateIdForRow()
* hasId()

Also stubbed out the rest of the tests for the row file
image


andrewc5716 (approved):
nice, looks good

PR #22, opened 2024-11-15
Adds logic to load the members and couples table from the Onestop

jbwong05:
This PR adds logic to load a members and couples table from the Onestop into the script properties. The table is expected to be on the `Members` and `Couples` tabs respectively and should be of the following formats
Screenshot 2024-11-14 at 6 37 19 PM
Screenshot 2024-11-14 at 6 39 10 PM
The logic read these tables and load 2 maps into script properties: 
1. A MEMBER_MAP which maps a members name to their properties (ex. Josh Wong -> {name: Josh Wong, gender: Male, married: false, parent: false, class: 2022}
3. An ALIAS_MAP which maps an alias to an array of member names the alias refers to. This alias can be either a couple's alias or a person's alternate name (ex. Josh W. -> [Josh Wong] or Andrew/Janice -> [Andrew Chan, Janice Chan] )

Test coverage:
Screenshot 2024-11-14 at 6 37 04 PM
In order to have different mocks for each test because each test may have different mocking requirements, the function/module under test has to be loaded/imported dynamically with `require` after all mocks for the test are setup. If the function under test is imported at the top of the file, it will be imported without any of the mocks for the current test being set which is not what we want

### DIFF
--- a/__test__/members.test.ts
+++ b/__test__/members.test.ts
@@ -1,0 +1,346 @@
+// Mock PropertiesService and SpreadsheetApp must be imported and set before imported modules require these global objects
+import { PropertiesService, SpreadsheetApp } from 'gasmask';
+global.PropertiesService = PropertiesService;
+global.SpreadsheetApp = SpreadsheetApp;
+
+import { getRandomlyGeneratedAliasMap, getRandomlyGeneratedAliasTable, getRandomlyGeneratedMemberMap, getRandomlyGeneratedMemberTable, getRandomlyGeneratedRange, getRandomlyGeneratedSheet, Mock } from './testUtils';
+import { TabNotFoundError } from '../src/main/error/tabNotFoundError';
+
+const NAME_COLUMN_INDEX: number = 0;
+const GENDER_COLUMN_INDEX: number = 1;
+const MARRIED_COLUMN_INDEX: number = 2;
+const PARENT_COLUMN_INDEX: number = 3;
+const CLASS_COLUMN_INDEX: number = 4;
+const ALTERNATE_NAMES_COLUMN_INDEX: number = 5;
+
+describe("MEMBER_MAP", () => {
+    it("should return the member map from the script properties when it is present", () => {
+        const memberMapMock: MemberMap = getRandomlyGeneratedMemberMap();
+
+        jest.doMock("../src/main/propertiesService", () => ({
+            getScriptProperty: jest.fn(() => JSON.stringify(memberMapMock)),
+        }));
+        
+        // Import the MEMBER_MAP with the mocked propertiesService
+        const { MEMBER_MAP } = require('../src/main/members');
+
+        expect(MEMBER_MAP).toEqual(memberMapMock);
+    });
+
+    it("should return an empty map when there is no member map present in the script properties", () => {
+        jest.doMock("../src/main/propertiesService", () => ({
+            getScriptProperty: jest.fn(() => null),
+        }));
+
+        // Import the MEMBER_MAP with the mocked propertiesService
+        const { MEMBER_MAP } = require('../src/main/members');
+
+        expect(MEMBER_MAP).toStrictEqual({});
+    });
+});
+
+describe("ALIASES_MAP", () => {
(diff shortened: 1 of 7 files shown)


src/main/members.ts
```diff
@@ -0,0 +1,126 @@
...
+    // Start at row 1 to skip the table header row
+    for(let i = 1; i < cellValues.length; i++) {
+        const rowValues: any[] = cellValues[i];
+        const currentMember: Member = constructMember(rowValues);
+        memberMap[currentMember.name] = currentMember;
+
+        const alternateNames: string[] = getAliasList(rowValues, ALTERNATE_NAMES_COLUMN_INDEX);
+        // Maps a person's alternate name to their actual name
+        alternateNames.forEach(alternateName => {
+            if(alternateNamesMap.hasOwnProperty(alternateName)) {
```
andrewc5716, line 62: can we add a warning log here? ideally we shouldn't be allowing this, so we should see when this is the case

src/main/members.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    return aliasMap;
+}
+
+function mergeAliasMaps(firstAliasMap: AliasMap, secondAliasMap: AliasMap): AliasMap {
+    const finalAliasMap: AliasMap = firstAliasMap;
+
+    const aliases: string[] = Object.keys(secondAliasMap);
+    for(const alias of aliases) {
+        if(finalAliasMap.hasOwnProperty(alias)) {
```
andrewc5716, line 113: can we also add a warning log for this? it would suck when this happens

__test__/members.test.ts
```diff
@@ -0,0 +1,346 @@
...
+    it("should map an alternate name to multiple people when multiple people share the same alternate name", () => {
+        const membersDataValuesMock: any[][] = getRandomlyGeneratedMemberTable(2, 1);
+        const membersDataRangeMock: Range = getRandomlyGeneratedRange();
+        membersDataRangeMock.getValues = jest.fn(() => membersDataValuesMock);
+        const membersTableSheetMock: Sheet = getRandomlyGeneratedSheet();
+        membersTableSheetMock.getName = jest.fn(() => "Members");
+        membersTableSheetMock.getDataRange = jest.fn(() => membersDataRangeMock);
+
+        membersDataValuesMock[1][NAME_COLUMN_INDEX] = "name1";
+        membersDataValuesMock[1][ALTERNATE_NAMES_COLUMN_INDEX] = "alias1,alias2";
```
andrewc5716, line 261: for tests like these, sometimes it helps to use more "real" looking values to make it easier to read, e.g. "Kevin L., Kevin", "Kevin H, Kevin". a little bit faster to read and tell where the duplication is

src/main/members.ts
```diff
@@ -0,0 +1,126 @@
...
+function loadMembersFromOnestop(): { memberMap: MemberMap, alternateNamesMap: AliasMap } {
+    const membersTab: Sheet = getTab(MEMBERS_TAB_NAME);
+    const dataRange: Range = membersTab.getDataRange();
+    const cellValues: any[][] = dataRange.getValues();
+
+    const memberMap: MemberMap = {};
+    const alternateNamesMap: AliasMap = {};
+
+    // Start at row 1 to skip the table header row
+    for(let i = 1; i < cellValues.length; i++) {
```
andrewc5716, line 54: if you feel like it, feel free to add an additional smaller function to break this up, would help a bit with reading this faster. technically there's triple nested blocks in here haha

andrewc5716 (commented):
on the whole, looks good. thanks for churning this out!

src/main/members.ts
jbwong05: Warning log added

src/main/members.ts
jbwong05: Warning log added here too

__test__/members.test.ts
jbwong05: Names and aliases used in the tests replaced with more realistic names generated by chatgpt. Didn't want to use our actual names for privacy sake

andrewc5716 (approved):
lgtm 👍

PR #24, opened 2024-11-22
Add Group and Supergroup loading

jbwong05:
This PR adds group and supergroup loading from the Onestop into script properties. After the `loadGroupsFromOnestopIntoScriptProperties()` function is run, a json map is created that maps a group name to an array of all of its group members
```
{
    "IGSM": ["personName1", "personName2", ...],
    "IUSM": ["personName3", "personName4", ...],
    "IUSM": ["personName1", "personName2", "personName3", "personName4", ...], // All names from IGSM and IUSM
}
```
Onestop tables:
Screenshot 2024-11-20 at 10 46 38 PM
Screenshot 2024-11-20 at 10 46 45 PM

Logic:
Loading groups with members is pretty straightforward. Just add them to the map. Loading Supergroups (groups of groups) is more complicated. This is actually my second implementation. My first used a naive approach which worked but was probably closer to O(n^2). It can be found here - https://github.com/andrewc5716/onestop-basecamp-integration/pull/23

Here is the logic I implemented for loading supergroups for this attempt. Algorithm is closer to DFS using maps for lookup which should be closer to O(n)
1. Load all Supergroups from the Onestop table into a map for lookup
2. Go through each of the Supergroups
* If all of that Supergroups dependent subgroups have already been loaded, then load the Supergroup
* Otherwise go through each of the dependent subgroups and recursively repeating this process to load the subgroup if it is not already loaded. Afterwards, all of the dependent subgroups should be loaded so the Supergroup can then be loaded

This PR admittedly also has a lot of object copying in an effort to avoid using output parameters

Changes: 
* `getCellValues()` and `loadMapFromScriptProperties()` were refactored so they can be used by both member and group loading. Unit tests were added for these refactored functions
* group loading and super group loading

Unit test report:
Screenshot 2024-11-21 at 5 39 03 PM

Successful execution on GAS:
Screenshot 2024-11-21 at 5 50 23 PM

### DIFF
--- a/__test__/groups.test.ts
+++ b/__test__/groups.test.ts
@@ -329,8 +329,8 @@ describe("loadGroupsFromOnestopIntoScriptProperties", () => {
             SDSU: ["Charlie Davis", "Emily Clark"],
             A2K: ["Frank Miller", "Grace Wilson"],
             Community: ["Frank Miller", "Grace Wilson"],
-            UCSD: ["John Doe", "Jane Smith", "Alice Johnson", "Bob Brown"],
             College: ["John Doe", "Jane Smith", "Alice Johnson", "Bob Brown", "Charlie Davis", "Emily Clark"],
+            UCSD: ["John Doe", "Jane Smith", "Alice Johnson", "Bob Brown"],
         };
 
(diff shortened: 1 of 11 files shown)


jbwong05:
Closing as #25 was merged

PR #25, opened 2024-11-24
Group and Iterative Supergroup loading

jbwong05:
Mirror of https://github.com/andrewc5716/onestop-basecamp-integration/pull/24 but with DFS implemented iteratively rather than recursively. In my opinion it is a little easier to read

### DIFF
--- a/src/main/groups.ts
+++ b/src/main/groups.ts
@@ -27,7 +27,6 @@ export function loadGroupsFromOnestopIntoScriptProperties(): void {
     const combinedGroupsMaps: GroupsMap = mergeGroupsMaps(groupsMap, supergroupsMap);
     const duplicateFreeGroupsMap: GroupsMap = removeDuplicatesFromGroupMaps(combinedGroupsMaps);
 
-    Logger.log(JSON.stringify(duplicateFreeGroupsMap));
     setScriptProperty(GROUPS_MAP_KEY, JSON.stringify(duplicateFreeGroupsMap));
 }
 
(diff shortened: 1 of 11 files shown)


src/main/groups.ts
```diff
@@ -0,0 +1,201 @@
...
+            const subgroups: string[] = currentSupergroup.subgroups;
+            for(const subgroup of subgroups) {
+                if(!isValidGroup(subgroup, allSupergroups, loadedGroups)) {
+                    // Skip invalid groups
+                    continue;
+                } else if(isSuperGroup(subgroup, allSupergroups) && !loadedSupergroups.hasOwnProperty(currentSupergroupName)) {
+                    // Add subgroups that are Supergroups that have not yet been loaded
+                    supergroupStack.push(subgroup);
+                }
+            }
```
jbwong05, line 103: This section can be refactored so we don't have a nested loop within this function but it would require having the stack as an output parameter or have the refactored function return a list of subgroups to add to the stack and then use list concatenation to actually add them. Would appreciate some input here

andrewc5716: i think the refactor makes sense, a list concat isn't ugly so i don't see much of a readability cost, or am i missing something?

andrewc5716:
you're right, this is way easier to read than the other one. i struggled on the other pr but was able to get thru this one on my first read

src/main/groups.ts
```diff
@@ -0,0 +1,201 @@
...
+        const groupName: string = rowValues[GROUP_NAME_COLUMN_INDEX];
+        const groupMemberNames: string[] = getGroupMemberNames(rowValues);
+
+        if(groupName === "") {
+            // Skip empty entries
+            continue;
+        }
+
+        if(groupsMap.hasOwnProperty(groupName)) {
+            Logger.log(`Group ${groupName} has already been defined in the Groups table. Combining the two lists of members`);
```
andrewc5716, line 53: maybe we can prefix these kind of logs with something like "WARN:", since we don't have log levels available for us

jbwong05: Added the `WARN` prefix

src/main/groups.ts
jbwong05: Refactored with a list concat

andrewc5716 (approved):
looks better with the refactor!

PR #26, opened 2024-12-02
Added system to filter group members

jbwong05:
Added a system to be able to filter groups of members based on Andrew's comment here - https://3.basecamp.com/4474129/buckets/38736474/todos/7962234197#__recording_7963320501. Should hopefully be pretty easy to add new filters in the future if we need them. Comments on how to add new filters have been left in the code. 

At the moment, if multiple filters are applied like `Bros` and `Married`, it will combine the filters and filter out anyone that is not a Bro and is not married (the result is bros that are married). If we want to make it so that two filters like `Bros` and `Married` result in bros OR anyone that is married, it is just a 1 line change changing `every()` to `some()` on this line here - https://github.com/andrewc5716/onestop-basecamp-integration/blob/7f1ce5e9722b774c7b86b87e7c34cedc2edb6a64/src/main/filter.ts#L49

### DIFF
--- a/src/main/filter.ts
+++ b/src/main/filter.ts
@@ -17,6 +17,8 @@ const FILTER_MAP: FilterMap = {
     Parents: parentsFilter,
     Moms: momsFilter,
     Dads: dadsFilter,
+    "Minus Moms": minusMomsFilter,
+    "Minus Dads": minusDadsFilter,
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/filter.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { MEMBER_MAP } from "./members";
+
+type FilterMap = { [onestopFilterName: string]: FilterFunction };
+
+const COMMA_DELIMITER: string = ",";
+const BROS_GENDER: string = "Male";
+const SIS_GENDER: string = "Female";
+// Maps a filter's name on the Onestop to its corresponding filter function
+const FILTER_MAP: FilterMap = {
```
andrewc5716, line 13: can we also add "minus moms" and "minus dads"? those are used for mbs every week

jbwong05: Added the `Minus Moms` and `Minus Dads` filters. Was pretty straightforward so hopefully adding additional filters shouldn't be too complicated

andrewc5716 (approved):

andrewc5716 (approved):

PR #27, opened 2024-12-06
Update to match current onestop format (+ domain column)

andrewc5716:
just adds domain column, removes food lead column


jbwong05 (commented):
Should we also update the `Who` column as it seems to now be the `Ministry` column?

andrewc5716:
> Should we also update the `Who` column as it seems to now be the `Ministry` column?

i was wondering about that too - i think "ministry" actually is inaccurate as a column name, it's more like "group", since daytime sisters for example, isn't a ministry. i was planning to bring it up to audrey and the mgmt team, until then i left it as is.

jbwong05 (approved):

PR #29, opened 2024-12-20
Fix unit tests

jbwong05:


__test__/groups.test.ts
```diff
@@ -55,7 +54,7 @@ describe("loadGroupsFromOnestopIntoScriptProperties", () => {
 
         const setScriptPropertyMock: Mock = jest.fn();
         jest.mock("../src/main/propertiesService", () => ({
-            loadMapFromScriptProperties: jest.fn(),
```
jbwong05, line 58: Because this function is getting mocked with `jest.fn()` which returns undefined, the following line was crashing because `GROUPS_MAP` was undefined
https://github.com/andrewc5716/onestop-basecamp-integration/blob/329b317aaf082b64c92d92153c1abbecac4dafe8/src/main/groups.ts#L29

__test__/groups.test.ts
```diff
@@ -2,7 +2,6 @@ import { Logger } from "gasmask";
 global.Logger = Logger;
 
 import { getRandomlyGeneratedGroupsMap, getRandomlyGeneratedGroupsTable, getRandomlyGeneratedSupergroupsTable, Mock } from "./testUtils";
-import { getMembersFromGroups } from "../src/main/groups";
```
jbwong05, line 5: Importing the function to test at the top of the file means that any mocking does not apply to this version of the function that was imported and is eventually called. This happens because JS is interpreted and the file is parsed top to bottom meaning if we are doing any mocking, it must be setup first and then the function/module being tested has to be imported with the mocks that were setup

__test__/groups.test.ts
```diff
@@ -498,34 +497,48 @@ describe('getMembersFromGroups', () => {
         const groupNames = ['SDSU', 'IUSM'];
 
         // Mock GROUPS_MAP in your test environment if needed
-        jest.mock('../src/main/groups', () => ({
+        jest.mock('../src/main/propertiesService', () => ({
             loadMapFromScriptProperties: jest.fn().mockReturnValueOnce(MOCK_GROUPS_MAP)
```
jbwong05, line 501: `loadMapFromScriptProperties` is part of the `propertiesService` module

__test__/groups.test.ts
```diff
@@ -498,34 +497,48 @@ describe('getMembersFromGroups', () => {
...
         }));
 
+        const { getMembersFromGroups } = require("../src/main/groups");
+
```
jbwong05, line 505: Because we are using dynamic module mocking, the function being tested has to be dynamically imported after all mocks for this test have been setup

src/main/row.ts
```diff
@@ -3,7 +3,7 @@ import { InvalidHashError } from "./error/invalidHashError";
 import { RowBasecampMappingMissingError } from "./error/rowBasecampMappingMissingError";
 import { RowMissingIdError } from "./error/rowMissingIdError";
 import { RowNotSavedError } from "./error/rowNotSavedError";
-import { FILTER_NAMES, filterMembers, getFilterNames } from "./filter";
```
jbwong05, line 6: This was throwing a compiler error because the function didn't exist causing the tests to freak out

src/main/row.ts
```diff
@@ -517,7 +517,7 @@ function filterMinistryAttendees(ministryNames: string[], ministryFilters: strin
  */
 function filterDomainAttendees(domainNames: string[], domainFilters: string[]): string[] {
     if (domainNames[0] === "Rotation") {
-        return getRotationMembers(); // TODO: Josh will implement this.
```
jbwong05, line 520: This was throwing a compiler error because the function didn't exist causing the tests to freak out

PR #30, opened 2024-12-21
basecamp events (schedule entries) api

andrewc5716:
adds create, update, delete functions for basecamp events (schedule entries)

PR #31, opened 2024-12-21
Unit test updates for #30

jbwong05:
Updates to the unit tests for #30 that brings the unit test style more inline with existing unit tests


andrewc5716 (approved):
@jbwong05 thank you!

PR #32, opened 2024-12-21
Updated helper column parsing

jbwong05:
## Changes
* filtering now requires an array of filter strings instead of a string of comma separated filters
* helper logic has been updated

## Helper logic updates
Note: `/` will be reserved for couple aliases

The existing parsing of the helper column has been updated to the following (Note: The order of the helpers doesn't matter for the following examples):
## Expected format
 ```
Role: helperName1, helperName2, group name, alias name, group with filter
Example: 
Food: John Doe, Jane Smith, UCSD Sis, Bob/Alice
Tech: Charlie Davis, Emily Wilson, SDSU
```
## Logic
1. Split the helper column by new lines. Each role should be on a separate line (this logic already exists and is not changed)
The following steps will be done for each line
2. Extract the role using the colon as the delimiter (this logic already exists and is not changed)
3. Separate each helper using the comma
4. For each individual helper that was comma separated do the following
* Extract any filters applied to this individual helper
* Check if the helper is a group. If so expand the group to all its members
* Check if the helper is an alias (such as for a couple). If so expand the alias
* Apply any individual filters that were detected
5. After all individual helpers have been processed, apply any overall filters that were detected to all helpers of this role
6. Remove any duplicate members that came up during group expansion

### CODE REVIEW

File: __test__/row.test.ts [hunk 1]
```diff
@@ -172,6 +173,435 @@ describe("getBasecampTodosForHelpers", () => {
...
+        ];
+
+        const { getHelperGroups } = require("../src/main/row");
+
+        const helperGroups: HelperGroup[] = getHelperGroups(rowMock);
+
+        expect(helperGroups).toStrictEqual(expectedHelperGroups);
+    });
+
+    it("should apply filters to all members of a HelperGroup when a filter is included in the list of helpers", () => {
```
jkzang, line 236: Hmm, do we need to have filters that apply to the entire list of helpers for a role? I feel like filters should only be used in conjunction with an existing group or alias instead of to the entire list of helpers (as you already implemented). 

Personally, I don't think people would input sisters as part of the helpers and then go back to add a filter to remove them and that might cause confusion with people being unintentionally filtered out although in the cell it mentions sisters. 

Was there a scenario that you had in mind in which this would be used?

jkzang (changes requested):

__test__/row.test.ts
jbwong05: @jkzang I think this was something Andrew mentioned unless I might have misunderstood. I guess it would be in the case where let's say sisters from UCSD, A2K, and IGSM were helping with something. Instead of specifying UCSD Sis, A2K Sis, IGSM Sis, you can just specify UCSD, A2K, IGSM, Sis. I see your point though

andrewc5716: @jkzang thanks for bringing this up, that's a good point. 
i think that for the helper column, since it's a hard rule to have everything be comma delimited, intuitively i think ppl would not assume a filter would apply past its comma limits. so let's not have filters apply to the whole list, as jack said.
however, it's true that the logic of having a filter apply to the whole list is what we'll be doing in the Who column. unfortunately this is inconsistent, it would've been nice to have the parsing logic be identical. in part, we don't rly have a choice since it's just one column and the multi-dropdown forces this, but at least it's still somewhat intuitive as ppl don't see commas, even though that's what the text ends up being when we parse it.

i may change my mind on this, but for now this is my thinking. lmk what you guys think.

andrewc5716 (approved):
aside from jack's comment, lgtm. technically this could go thru; most ppl won't put just a filter in the comma separated list and it should still "work", but to be air-tight might be best to take out that logic so there aren't any surprises.

__test__/row.test.ts
jbwong05: This makes sense especially to avoid confusion even if it's inconsistent. I'll remove the support for the filters in the comma separated filter list

jkzang (approved):

PR #33, opened 2025-01-05
Dev/determine attendees for an event

jkzang:


.clasp.json
```diff
@@ -1,4 +1,4 @@
 {
-  "scriptId": "1N22pZ29WY4DF-KV-JZ7jLh92f44pWrgL0rUqLviBbQ8ECZCEMbpvxi9V",
+  "scriptId": "1wjCZWISevP9KgxxiNM9OPP9Aq3VnGIxVfS6BTj586WhfrKWTg8rRmvgq",
```
andrewc5716, line 2: this should be ignored

__test__/row.test.ts
```diff
@@ -99,6 +99,193 @@ describe("generateIdForRow", () => {
...
+
+                } else if (key === "GROUPS_MAP") {
+                    return MOCK_GROUPS_MAP;
+                } 
+            }),
+        }));
+
+        const { getAttendeesFromRow } = require("../src/main/row");
+
+        expect(() => getAttendeesFromRow(row)).toThrow(
```
andrewc5716, line 203: for now let's go with silent errors; i saw even today there's "no mbs" in a what column and nothing else filled out

src/main/row.ts
```diff
@@ -455,6 +472,199 @@ function getHelperGroupFromHelperList(helperNameList: string, role: string | und
...
+/**
+ * Process attendees for domains.
+ * 
+ * @param domainNames - Names of domains.
+ * @param domainFilters - Filters to apply to the domain members.
+ * @returns Array of filtered members.
+ */
+function filterDomainAttendees(domainNames: string[], domainFilters: string[]): string[] {
+    if (domainNames[0] === "Rotation") {
+        return []; // TODO: Josh will implement this.
```
andrewc5716, line 646: this should now be implemented?

src/main/row.ts
```diff
@@ -455,6 +472,199 @@ function getHelperGroupFromHelperList(helperNameList: string, role: string | und
...
+        attendees.push(...getAllHelperNames(row));
+        
+    } else if(ministryNames.length > 0) {
+        // Step 3: Process Ministry Attendees
+        const ministryAttendees = filterMinistryAttendees(ministryNames, ministryFilters);
+        attendees.push(...ministryAttendees);
+
+    } else if(domainNames.length > 0) {
+        // Step 4: Process Domain Attendees
+        const domainAttendees = filterDomainAttendees(domainNames, domainFilters);
```
andrewc5716, line 615: i can foresee a use case where the domain will be "churchwide", and ministry/who will be "sisters". i think we should be able to handle this case where we need to use the filter in the ministry/who column and apply it to the domain

.clasp.json
jkzang: I think it was checked in a while back which means it can't be ignored. I'll go ahead and delete the file from source control and I think after that the .clasp.json entry in the .gitignore will take effect

src/main/row.ts
jkzang: Yup, you're right

__test__/row.test.ts
jkzang: Sounds good, I replaced the error with a console.log

src/main/row.ts
jkzang: Done

src/main/row.ts
```diff
@@ -455,6 +472,213 @@ function getHelperGroupFromHelperList(helperNameList: string, role: string | und
...
+        // Process Ministry Attendees
+        const ministryAttendees = filterMinistryAttendees(ministryNames, ministryFilters);
+        attendees.push(...ministryAttendees);
+
+    } else if(domainNames.length > 0 && ministryFilters.length === 0) {
+        // Process Domain Attendees
+        const domainAttendees = filterDomainAttendees(domainNames, domainFilters);
+        attendees.push(...domainAttendees);
+
+    } else if(isChurchwide && ministryFilters.length > 0) {
```
andrewc5716, line 629: sorry i should've been more clear; i think this should apply for any domain, not just churchwide, e.g. "college" as domain and "sisters" as group should also work.

__test__/row.test.ts
```diff
@@ -99,6 +99,219 @@ describe("generateIdForRow", () => {
...
+        }));
+
+        const { getAttendeesFromRow } = require("../src/main/row");
+
+        const result = getAttendeesFromRow(row);
+
+        expect(result).toEqual(['Brian Lin', 'James Lee', 'Jack Zhang', 'Angel Zhang']);
+    });
+
+    // it('should not throw an error when both ministry and domain names are missing', () => {
```
andrewc5716, line 183: this should be uncommented out, and the test can just be edited so that it makes sure no error is thrown

__test__/row.test.ts
```diff
@@ -99,6 +99,219 @@ describe("generateIdForRow", () => {
     });
 });
 
+
+
+// Mocked GROUPS_MAP
+const MOCK_GROUPS_MAP = {
```
andrewc5716, line 105: nitpick; in order to read the unit tests, i have to look back and forth between these defined mock objects and the tests. though it'd be repetitive, it'd make reading easier if a slimmed down object is in the unit test itself

src/main/row.ts
jkzang: Ok I got it. I modified the logic. Let me know if this is more in line with what you were thinking.

__test__/row.test.ts
jkzang: Ok sounds good, I uncommented and modified the test

__test__/row.test.ts
jkzang: K I moved all the mock data into the tests

src/main/row.ts
```diff
@@ -455,6 +472,202 @@ function getHelperGroupFromHelperList(helperNameList: string, role: string | und
...
+        const ministryAttendees = filterMinistryAttendees(ministryNames, ministryFilters);
+        attendees.push(...ministryAttendees);
+
+    } else if(domainNames.length > 0 && ministryFilters.length === 0) {
+        // Process Domain Attendees
+        const domainAttendees = filterDomainAttendees(domainNames, domainFilters);
+        attendees.push(...domainAttendees);
+
+    } else if(domainNames.length > 0 && ministryFilters.length > 0) {
+        // Special case: e.g. If CHURCHWIDE is the domain and only a filter is set in the ministry column
```
andrewc5716, line 619: lingering comment needs updating

__test__/row.test.ts
```diff
@@ -99,6 +99,245 @@ describe("generateIdForRow", () => {
...
+        }));
+
+        const { getAttendeesFromRow } = require("../src/main/row");
+
+        const result = getAttendeesFromRow(row);
+
+        expect(result).toEqual(['Janice Chan', 'Joyce Lai']);
+    });
+
+    it('should apply any filters present in the ministry column to the CHURCHWIDE group if domain value is CHURCHWIDE', () => {
```
andrewc5716, line 269: test case name needs to be updated

.clasp.json
jbwong05, line 1: I don't think we should delete this file or else `npm run deploy:dev` and `npm run deploy:prod` will fail

src/main/filter.ts
```diff
@@ -8,8 +8,9 @@ type FilterMap = { [onestopFilterName: string]: FilterFunction };
 
 const BROS_GENDER: string = "Male";
 const SIS_GENDER: string = "Female";
+
 // Maps a filter's name on the Onestop to its corresponding filter function
-const FILTER_MAP: FilterMap = {
+export const FILTER_MAP: FilterMap = {
```
jbwong05, line 13:
```suggestion
const FILTER_MAP: FilterMap = {
```
This isn't being used externally for the sake of abstraction so we don't need to export this

src/main/row.ts
```diff
@@ -357,6 +358,22 @@ export function getBasecampTodosForHelpers(row: Row): RoleRequestMap {
...
+ */
+export function getHelperLines(row: Row): string[] {
+    return row.helpers.value.split(NEW_LINE_DELIM);
+}
+
+/**
+ * @param helperLine a line from the helpers column in a row
+ * @returns true or false
+ */
+export function hasRole(helperLine: string) {
```
jbwong05, line 373:
```suggestion
export function hasRole(helperLine: string): boolean {
```

src/main/row.ts
```diff
@@ -455,6 +472,202 @@ function getHelperGroupFromHelperList(helperNameList: string, role: string | und
...
+
+/**
+ * Checks if the event has a who (ministry) column value of OTHER
+ * 
+ * @param row - An event row
+ * @returns true or false
+ */
+export function CheckForOther(row: Row): boolean {
+    return splitWhoColumn(row)[0].toUpperCase() ===  "OTHER";
+}
```
jbwong05, line 581:
```suggestion
export function checkForRotation(row: Row): boolean {
    return splitWhoColumn(row)[0].toUpperCase() === "ROTATION";
}

/**
 * Checks if the event has a who (ministry) column value of VARIOUS
 * 
 * @param row - An event row
 * @returns true or false
 */
export function checkForVarious(row: Row): boolean {
    return splitWhoColumn(row)[0].toUpperCase() ===  "VARIOUS";
}

/**
 * Checks if the event has a who (ministry) column value of OTHER
 * 
 * @param row - An event row
 * @returns true or false
 */
export function checkForOther(row: Row): boolean {
    return splitWhoColumn(row)[0].toUpperCase() ===  "OTHER";
}
```
Can we use camel case for these?

src/main/row.ts
```diff
@@ -455,6 +472,202 @@ function getHelperGroupFromHelperList(helperNameList: string, role: string | und
...
+ * @param domainNames - Names of domains.
+ * @param ministryNames - Names of ministry groups.
+ */
+function handleMissingData(domainNames: string[], ministryNames: string[]): void {
+    if (domainNames.length === 0 && ministryNames.length === 0) {
+        throw new DomainMissingError(
+            "ERROR: Unable to get attendees from row becuase both domain and ministry columns are empty!"
+        );
+    }
+}
```
jbwong05, line 668: It seems like this function is not being used. Is it needed? Looks like it is commented out on line 626

jbwong05 (approved):

PR #34, opened 2025-01-05
add helper cell validation

andrewc5716:
create a function for validating a helper cell. unfortunately this function can't be used from data > data validation > custom formula. it has to be indirectly used, so in a hidden column, e.g. H, run `isHelperCellValid(A1)`, then in data validation, do `H1=TRUE`.
currently it's named to be used for the helper cell, but honestly this can just be used for the lead cell too. technically a group can't be put in there, but the code is almost identical so i'm ok with shortcutting this for now?

### DIFF
--- a/src/main/filter.ts
+++ b/src/main/filter.ts
@@ -9,7 +9,7 @@ type FilterMap = { [onestopFilterName: string]: FilterFunction };
 const BROS_GENDER: string = "Male";
 const SIS_GENDER: string = "Female";
 // Maps a filter's name on the Onestop to its corresponding filter function
-export const FILTER_MAP: FilterMap = {
+const FILTER_MAP: FilterMap = {
     Bros: brosFilter, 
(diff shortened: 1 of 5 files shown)


src/main/filter.ts
```diff
@@ -9,7 +9,7 @@ type FilterMap = { [onestopFilterName: string]: FilterFunction };
 const BROS_GENDER: string = "Male";
 const SIS_GENDER: string = "Female";
 // Maps a filter's name on the Onestop to its corresponding filter function
-const FILTER_MAP: FilterMap = {
+export const FILTER_MAP: FilterMap = {
```
jbwong05, line 12:
```suggestion
const FILTER_MAP: FilterMap = {
```
Does not look like this is being used outside of this module so it does need to be exported

jbwong05 (approved):

PR #35, opened 2025-01-17
Create update and delete events

jbwong05:
PR to create, update, and delete events

### DIFF
--- a/__test__/main.test.ts
+++ b/__test__/main.test.ts
@@ -35,7 +35,6 @@ describe("importOnestopToBasecamp", () => {
             .mockReturnValueOnce(true)
             .mockReturnValueOnce(false)
             .mockReturnValueOnce(true);
-        const hasChangedMock: Mock = jest.fn();
         const getBasecampTodoRequestsForRowMock: Mock = jest.fn()
             .mockReturnValueOnce(roleRequestMapMock1)
             .mockReturnValueOnce(roleRequestMapMock2);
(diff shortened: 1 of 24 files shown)


src/main/main.ts
```diff
@@ -57,7 +59,13 @@ function processExistingRow(row: Row): void {
 
         const updatedRoleTodoIdMap: RoleTodoIdMap = {...existingRoleTodoIdMap, ...newRoleTodoIdMap};
 
-        saveRow(row, updatedRoleTodoIdMap);
+        // Handles Schedule Entries
```
andrewc5716, line 62: small thing, but now looks like the todo and schedule entry portions of this function could be split into 2 functions

__test__/main.test.ts
```diff
@@ -0,0 +1,615 @@
...
+    
+        // Asserts for new second row
+        expect(hasIdMock).toHaveBeenNthCalledWith(3, rowMock2);
+        expect(getBasecampTodoRequestsForRowMock).toHaveBeenNthCalledWith(2, rowMock2);
+        expect(createNewTodosMock).toHaveBeenNthCalledWith(2, roleRequestMapMock2);
+        expect(getScheduleEntryRequestForRowMock).toHaveBeenNthCalledWith(2, rowMock2);
+        expect(createScheduleEntryMock).toHaveBeenNthCalledWith(2, scheduleEntryRequestMock2, scheduleIdentifierMock);
+        expect(generateIdForRowMock).toHaveBeenNthCalledWith(2, rowMock2);
+        expect(saveRowMock).toHaveBeenNthCalledWith(2, rowMock2, roleTodoIdMapMock2, scheduleEntryIdMock2);
+        expect(hasIdMock).toHaveBeenNthCalledWith(4, rowMock2);
```
andrewc5716, line 129: wait, this is asserted to have been called 3x on line 122?

__test__/main.test.ts
```diff
@@ -0,0 +1,615 @@
...
+
+        // Asserts for new first row
+        expect(hasIdMock).toHaveBeenNthCalledWith(1, rowMock1);
+        expect(getBasecampTodoRequestsForRowMock).toHaveBeenNthCalledWith(1, rowMock1);
+        expect(createNewTodosMock).toHaveBeenNthCalledWith(1, roleRequestMapMock1);
+        expect(getScheduleEntryRequestForRowMock).toHaveBeenNthCalledWith(1, rowMock1);
+        expect(createScheduleEntryMock).toHaveBeenNthCalledWith(1, scheduleEntryRequestMock1, scheduleIdentifierMock);
+        expect(generateIdForRowMock).toHaveBeenNthCalledWith(1, rowMock1);
+        expect(saveRowMock).toHaveBeenNthCalledWith(1, rowMock1, roleTodoIdMapMock1, scheduleEntryIdMock1);
+        expect(hasIdMock).toHaveBeenNthCalledWith(2, rowMock1);
```
andrewc5716, line 118: generally we should aim for minimal assertions; i don't think we need to go as granular as checking for these functions like hasId, just the big ones like createNewTodos, createScheduleEntry, etc

andrewc5716: otherwise the unit tests get too tightly coupled with the implementation, and if i were to for whatever reason make an extra hasId call, this shouldn't necessarily fail

jbwong05: agreed

src/main/main.ts
jbwong05: Refactored

andrewc5716 (approved):
lgtm, thanks for your hard work!

PR #37, opened 2025-01-26
Hyperlink support

jbwong05:
Not fully unit tested but opening given the time constraints of this Monday because figuring out the unit testing may take a while. Adds support for hyperlinks and partial strikethrough of Text
Screenshot 2025-01-25 at 7 32 39 PM

### DIFF
--- a/src/main/row.ts
+++ b/src/main/row.ts
@@ -302,40 +302,44 @@ function getBasecampTodoDescription(row: Row): string {
     const startTime: string = row.startTime.toLocaleTimeString(locales, options);
     const endTime: string = row.endTime.toLocaleTimeString(locales, options);
 
-    const time: string = wrapWithPreTag(`WHEN: ${startTime} - ${endTime}`);
+    const time: string = `WHEN: ${startTime} - ${endTime}`;
     const inCharge: string = getRichTextFromText("IN CHARGE", row.inCharge);
     const helpers: string = getRichTextFromText("HELPERS", row.helpers);
     const notes: string = getRichTextFromText("NOTES", row.notes);
 
-    return wrapWithDivTag(location + time + inCharge + helpers + notes);
+    return wrapWithDivTag(combineWithBreakTags([location, time, inCharge, helpers, notes]));
 }
 
 function getRichTextFromText(prefix: string, text: Text): string {
     if(text.tokens.length === 1) {
-        return wrapWithPreTag(`${prefix}: ${text.value}`);
+        return `${prefix}: ${replaceNewLinesWithBreakTags(text.value)}`;
     }
 
     const textTokens: TextData[] = text.tokens;
-    let richText: string = "";
+    let richText: string = `${prefix}: `;
     for(const token of textTokens) {
         if(token.hyperlink !== null) {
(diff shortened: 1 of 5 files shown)


src/main/row.ts
```diff
@@ -343,8 +370,7 @@ export function getBasecampTodosForHelpers(row: Row): RoleRequestMap {
         const roleTitle: string = helperGroup.role ? `${helperGroup.role} Helper` : "Helper";
         const basecampTodoContent: string = `${roleTitle}: ${row.what.value}`;
         const basecampTodoDescription: string = getBasecampTodoDescription(row);
-        const helperIds: string[] = helperGroup.helperIds.filter(id => !leadIds.includes(id));
-        const assigneeIds: string[] = leadIds.concat(helperIds);
```
jbwong05, line 347: This seemed like a bug that resulted in leads getting assigned to the Todos for Helpers. Seems redundant so I removed the extra inclusion of leads on helpers Todos. Let me know if we should keep it the way it is

src/main/scan.ts
```diff
@@ -150,31 +149,31 @@ function getCellData(dataRange: Range): CellData[][] {
...
-            strikethrough: getCellStrikethrough(textStyle)
-        }
+        const newColumnData: CellData = {
+            value: cellValues[j],
+            richTextValue: cellRichTextData[j],
+        };
+        columnData.push(newColumnData);
     }
+
+    return columnData;
```
jbwong05, line 176: Refactor to fix output parameter

src/main/row.ts
andrewc5716: originally i had it that way so that leads would be able to easily find the todo associated with the helpers, in case they needed to communicate anything to them. but maybe it adds too much bloat, i'm fine to remove this and if ppl ask for it, we can add it back in

src/main/row.ts
```diff
@@ -303,12 +302,40 @@ function getBasecampTodoDescription(row: Row): string {
...
+
+    return wrapWithPreTag(richText);
+}
 
-    return location + time + inCharge + helpers + notes;
+function wrapWithDivTag(text: string): string {
+    return `<div>${text}</div>`;
+}
+
+function wrapWithPreTag(text: string): string {
```
andrewc5716, line 337: i think it's ok to not wrap with a pre tag, unless does something bad happen if we extract rich text from the date cell?

andrewc5716:
in your screenshot it looks like the "where" prefix is missing? but from the code it looked ok, so not sure if it's just the screenshot

jbwong05:
> in your screenshot it looks like the "where" prefix is missing? but from the code it looked ok, so not sure if it's just the screenshot

Good catch, this was actually a small bug

src/main/row.ts
jbwong05: @andrewc5716 The `pre` tag was used to preserve any of the formatting of the original cell. I have removed the wrapping `pre` tags and have now handled any line breaks using the `br` tag. This is the result:
Screenshot 2025-01-25 at 7 32 39 PM
Let me know what you think

andrewc5716 (approved):

PR #38, opened 2025-01-26
Group aliasing

jbwong05:
Adds group aliasing and refactors the way aliases are loaded since they are now shared by both members and groups. To perform all loading, we now just need to trigger `loadMembersAndGroupsFromOnestopIntoScriptProperties()`. Calling the individual member or group loading functions will leave out any aliases

Not all unit tests are done for this too, but was tested manually. Opening due to tight timeline

New format of tables with added alias column
Screenshot 2025-01-25 at 6 37 08 PM
Screenshot 2025-01-25 at 6 37 20 PM


andrewc5716 (approved):
nice, lgtm

PR #39, opened 2025-01-26
Allow blank lead/helper columns

andrewc5716:
There's now a use case where there's no one in charge/helping (churchwide prayer). The old logic would have the entire row not saved, but that would mean each time the script runs, a new schedule entry is created.


src/main/main.ts
```diff
@@ -89,10 +89,8 @@ function processNewRow(row: Row): void {
     const scheduleEntryRequest: BasecampScheduleEntryRequest = getScheduleEntryRequestForRow(row);
     const scheduleEntryId: string = createScheduleEntry(scheduleEntryRequest, getDefaultScheduleIdentifier());
 
-    if(Object.keys(roleTodoIdMap).length > 0 && scheduleEntryId !== "") {
-        generateIdForRow(row);
-        saveRow(row, roleTodoIdMap, scheduleEntryId);
-    }
```
jbwong05, line 95: Does this also introduce the case where we will be saving the save even if creating the Todo and Schedule Entry fail for some reason?

andrewc5716: i'm assuming that if either todo or schedule entry creation fails, the exception is thrown and we won't get to this point of the code?

jbwong05: Good point. We probably should be able to handle these kinds of exceptions gracefully

jbwong05 (approved):

PR #40, opened 2025-01-26
Refactored the isHelperCellValid method to return a helpful and descr…

jkzang:
I refactored the isHelperCellValid method to return a message containing a string of invalid helper identifiers instead of just false when the helper cell has invalid values. 

I started working on the validation for leads but haven't finished it.

I also added some additional unit test for the refactored isHelperCellValid method.

### DIFF
--- a/__test__/validation.test.ts
+++ b/__test__/validation.test.ts
@@ -1,14 +1,14 @@
 import { PropertiesService } from 'gasmask';
 global.PropertiesService = PropertiesService;
-import {isHelperCellValid} from "../src/main/validation";
+import {validateHelperCellText, validateLeadCellText} from "../src/main/validation";
 
-describe("isHelperCellValid", () => {
+describe("validateHelperCellText", () => {
     it("should allow empty", () => {
-        expect(isHelperCellValid("")).toBeTruthy();
+        expect(validateHelperCellText("")).toBe("");
     });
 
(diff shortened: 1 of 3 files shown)


jkzang:
I added all the logic for validating the leads cell and also integrated Andrew's normalization logic.

__test__/validation.test.ts
```diff
@@ -196,8 +212,187 @@ describe("isHelperCellValid", () => {
...
+                        "janice chan": {"gender": "Female"},
+                    }
+                } 
+                return {};
+            }),
+        }));
+        const { validateLeadCellText } = require("../src/main/validation");
 
-        expect(isHelperCellValid("SDSU staff")).toBeTruthy();
+        expect(validateLeadCellText("SDSU, Jack/Angel\nFood: Andrew Chan\nJanice Chan")).toBe("Invalid identifier(s): SDSU, Food: Andrew Chan");
```
andrewc5716, line 396: i'm not sure i understand, why is "Food: Andrew Chan" invalid here?

jkzang: It's because lead cells are only suppose to have names of people and not roles

jkzang: It's valid for the validateHelperCellText method

andrewc5716 (approved):

PR #42, opened 2025-01-27
Normalize all inputs

andrewc5716:
Normalize inputs for case insensitivity and with extra whitespace, parentheses removed. 

Additional unrelated changes/bug fixes:
- fix usage of `BASECAMP_PROJECT_ID` in people.ts, after change to how the configs are loaded
- no longer allow the same name alias for multiple ppl when loading from the onestop


src/main/row.ts
```diff
@@ -575,7 +577,8 @@ function getMinistryNames(row: Row): string[] {
  */
 function getMinistryFilters(row: Row): string[] {
     return splitWhoColumn(row)
-    .map(name => name.trim()).filter(value => isFilter(value));;
+    .map(name => name.toLowerCase().trim())
+    .filter(value => isFilter(value));;
```
jbwong05, line 581: Nit: Extra semicolon

jbwong05 (dismissed):

jbwong05 (approved):

PR #43, opened 2025-01-27
Todo links in event description

jbwong05:
* Renames `RoleTodoIdMap` to `RoleTodoMap` as we are no longer just mapping to just the Basecamp Todo id
* Adds corresponding Todo links to the Schedule Entry event description

Screenshot 2025-01-27 at 2 50 13 PM


andrewc5716 (approved):
nice, lgtm!

PR #44, opened 2025-01-28
Always add helpers as Schedule Entry attendees

jbwong05:
Always adds helpers as Schedule Entry attendees. More context here - https://3.basecamp.com/4474129/buckets/38736474/todos/8260556259


andrewc5716 (dismissed):
nice, glad this was a one line change haha. unfortunately you'll have to change the unit tests a bit when you merge from master, but otherwise looks good

andrewc5716 (approved):

PR #46, opened 2025-02-03
HTTP exception resiliency

jbwong05:
Adds resiliency for HTTP exceptions when calling the Basecamp API with the following:
* Use retries with exponential backoff (courtesy of ChatGPT)
  * Note: Unit testing for this is objectively not the best. I don't know how to unit test exponential backoff
  * 4xx errors are not retried
* All API calls wrapped with try/catch to not halt execution on an exception 
* If a row has not changed but it is missing Todos (implies at least one Todo was failed to be created last run), then run the existing logic to delete obsolete todos, update current ones, and create missing ones
* If a row has not changed and there are no missing todos but the schedule entry is missing, create the new schedule entry
* If at any point the schedule entry needs to be updated but it is missing, create the missing one

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -47,10 +47,12 @@ export function importOnestopToBasecamp(): void {
 function processExistingRow(row: Row): void {
 
     if(hasChanged(row) || isMissingTodos(row)) {
+        // Update Todos and Schedule Entry if the row has changed or if Todos are missing
         const updatedRoleTodoMap: RoleTodoMap = handleTodosForExistingRow(row);
         const scheduleEntryId: string = handleScheduleEntryForExistingRow(row, updatedRoleTodoMap);
         saveRow(row, updatedRoleTodoMap, scheduleEntryId);
     } else if(isMissingScheduleEntry(row)) {
+        // Create the Schedule Entry if it is missing
         const roleTodoMap: RoleTodoMap = getRoleTodoMap(row);
         const scheduleEntryId: string = handleScheduleEntryForExistingRow(row, roleTodoMap);
         saveRow(row, roleTodoMap, scheduleEntryId);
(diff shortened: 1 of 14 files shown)


src/main/schedule.ts
```diff
@@ -79,4 +80,16 @@ function getUpdateScheduleEntryUrl(scheduleEntryIdentifier: ScheduleEntryIdentif
 
 function getDeleteScheduleEntryUrl(scheduleEntryIdentifier: ScheduleEntryIdentifier): string {
     return getBasecampProjectUrl(scheduleEntryIdentifier.projectId) + RECORDINGS_PATH + scheduleEntryIdentifier.scheduleEntryId + TRASHED_STATUS_JSON_PATH;
-}
\ No newline at end of file
+}
+
+export function createScheduleEntryForRow(row: Row, roleTodoMap: RoleTodoMap): string {
+    const scheduleEntryRequest: BasecampScheduleEntryRequest = getScheduleEntryRequestForRow(row, roleTodoMap);
+    let scheduleEntryId: string = "";
```
andrewc5716, line 87: relying on the outside functions to know to check for `""` is more fragile, can this function instead return `string | undefined`?

src/main/main.ts
```diff
@@ -46,11 +46,16 @@ export function importOnestopToBasecamp(): void {
...
 function processExistingRow(row: Row): void {
 
-    if(hasChanged(row)) {
+    if(hasChanged(row) || isMissingTodos(row)) {
+        // Update Todos and Schedule Entry if the row has changed or if Todos are missing
         const updatedRoleTodoMap: RoleTodoMap = handleTodosForExistingRow(row);
         const scheduleEntryId: string = handleScheduleEntryForExistingRow(row, updatedRoleTodoMap);
-
         saveRow(row, updatedRoleTodoMap, scheduleEntryId);
+    } else if(isMissingScheduleEntry(row)) {
```
andrewc5716, line 54: can we log when we enter here? would be good to know if this is happening

andrewc5716 (approved):
nice, thank you!

PR #47, opened 2025-02-03
Dev/append helpful notes for todos

jkzang:
Hi @andrewc5716 my OneStop integration test environment is not setup properly at the moment. I wrote a unit test for this feature but as far as actually testing it to see if it works, I don't have the proper setup to do that at the moment. I will need to fix it eventually. 

Could you give it a test to see if the hyperlink appears in todos as well as let me know if there are any other points of improvement for this PR? 

Thanks!

### DIFF
--- a/__test__/row.test.ts
+++ b/__test__/row.test.ts
@@ -4,7 +4,7 @@ import { Logger, PropertiesService } from 'gasmask';
 global.Logger = Logger;
 global.PropertiesService = PropertiesService;
 
-import { generateIdForRow, getId, getMetadata, getSavedScheduleEntryId, hasId, saveRow } from "../src/main/row";
+import { generateIdForRow, generateOneStopHyperLinkHtml, getId, getMetadata, getSavedScheduleEntryId, hasId, saveRow } from "../src/main/row";
 import { RowMissingIdError } from '../src/main/error/rowMissingIdError';
 import { getRandomlyGeneratedByteArray, getRandomlyGeneratedMember, getRandomlyGeneratedMetadata, getRandomlyGeneratedRange, getRandomlyGeneratedRoleTodoMap, getRandomlyGeneratedRow, getRandomlyGeneratedRowBasecampMapping, getRandomlyGeneratedText, Mock } from './testUtils';
 import randomstring from "randomstring";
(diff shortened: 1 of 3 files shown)


__test__/row.test.ts
```diff
@@ -992,3 +992,13 @@ describe("getScheduleEntryRequestForRow", () => {
         expect(scheduleEntryRequest.participant_ids).toContain(PEOPLE_MAP["alice johnson"]);
     });
 });
+
+describe("constructOneStopHyperLink", () => {
+
+    it("should generate a valid hyperlink html snippet", () => {
+        const rowMock: Row = getRandomlyGeneratedRow();
+        const hyperlink = generateOneStopHyperLinkHtml(rowMock);
```
andrewc5716, line 1000: general rule of thumb for unit tests is that you should be able to read/understand the whole unit test without needing to look elsewhere. here the exact values of the link is hidden in this function, when it's being asserted on later. what should happen instead is you should show the hardcoding of the spreadsheet id and such here in this test, and then it'll match up with the expected hyperlink

andrewc5716:
@jkzang unfortunately the link doesn't show up. i think reading from metadata isn't working as expected. @jbwong05 you're more familiar with how the metadata works. when i tried playing around with metadata, seems that retrieving it doesn't get me anything useful.

jbwong05:
@andrewc5716 @jkzang Sorry, just seeing this now. I'll take a look

src/main/row.ts
```diff
@@ -285,6 +285,59 @@ function getLeadsNames(row: Row): string[] {
...
 
+/**
+ * Gets the spreadsheet id from a row
+ * 
+ * @param row row to extract spreadsheet id from
+ * @returns spreadsheet ID string or undefined
+ */
+
+function getSpreadSheetIdFromRow(row: Row): string | undefined {
+    return row.metadata.getLocation().getSpreadsheet()?.getId();
```
jbwong05, line 296:
```suggestion
    return row.metadata.getLocation().getRow()?.getSheet().getParent().getId();
```

src/main/row.ts
```diff
@@ -285,6 +285,59 @@ function getLeadsNames(row: Row): string[] {
...
+
+/**
+ * Gets the sheet id from a row, corresponding to the tab that a row is from
+ * 
+ * @param row row to extract sheet id from
+ * @returns spreadsheet ID string or undefined
+ */
+
+function getSheetIdFromRow(row: Row): string | undefined {
+    return row.metadata.getLocation().getSheet()?.getSheetId().toString();
```
jbwong05, line 307:
```suggestion
    return row.metadata.getLocation().getRow()?.getSheet().getSheetId().toString();
```

jbwong05:
@jkzang I believe the above suggestions should fix it. Tested it on my test spreadsheet. Issue was the Location object was a Row location instead of Spreadsheet location so calling `getSpreadsheet()` returns undefined but `getRow()` will give you the row and from there you can extract all of the Sheet data you need

PR #48, opened 2025-02-09
schedule entry summary always shows attendees and other fixes

andrewc5716:
currently the schedule entry summary/title shows "[]" if the attendee column is not a group, e.g. rotation, various, filter, etc.
quick and dirty way to fix this is to just display the whole attendee column.

also snuck in:
- adding "what" to the todo in richtext, since ppl put links in the "what" column
- always add leads/incharge to the attendee list, just in case. and simplifies logic a bit.


woenlee (approved):
LGTM

PR #53, opened 2025-02-10
Basecamp links in onestop

jbwong05:
PR to add Basecamp links into the Basecamp column of the Onestop
Screenshot 2025-02-10 at 10 22 23 AM

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -88,12 +88,8 @@ function handleScheduleEntryForExistingRow(row: Row, updatedRoleTodoMap: RoleTod
         }
     } else {
         // Create the Schedule Entry if it is missing
-        const basecampScheduleEntry: BasecampScheduleEntry | undefined = createScheduleEntryForRow(row, updatedRoleTodoMap);
+        const basecampScheduleEntry: BasecampScheduleEntry | undefined = createNewScheduleEntry(row, updatedRoleTodoMap);
         scheduleEntryId = basecampScheduleEntry?.id;
-
-        if(basecampScheduleEntry !== undefined) {
-            addBasecampLinkToRow(row, basecampScheduleEntry.url);
-        }
     }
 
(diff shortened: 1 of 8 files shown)


andrewc5716:
oh wow we can squeeze this in for rollout today, thank you!!
small request, would you be able to change the hyperlink text to something like "Link" so that it's shorter?

src/main/main.ts
```diff
@@ -87,7 +87,12 @@ function handleScheduleEntryForExistingRow(row: Row, updatedRoleTodoMap: RoleTod
         }
     } else {
         // Create the Schedule Entry if it is missing
-        scheduleEntryId = createScheduleEntryForRow(row, updatedRoleTodoMap);
+        const basecampScheduleEntry: BasecampScheduleEntry | undefined = createScheduleEntryForRow(row, updatedRoleTodoMap);
```
andrewc5716, line 90: looks like a little bit of code duplication, can this be separated into a function?

andrewc5716:
i'm assuming this doesn't affect `hasChanged(row)`given this happens before the row saves, did you happen to double check this in a quick test?

jbwong05:
@andrewc5716 This shouldn't affect `hasChanged(row)`. The row hash for `hasChanged` depends on the toString representation of the row object which does not include the Basecamp link column. We are not storing the Basecamp link column as part of the row during the scan phase so it should not affect the row hash and changed status at all
https://github.com/andrewc5716/onestop-basecamp-integration/blob/122a5a26780bddc0013f15ef80670d04daddb6cb/src/main/row.ts#L143
https://github.com/andrewc5716/onestop-basecamp-integration/blob/122a5a26780bddc0013f15ef80670d04daddb6cb/src/main/row.ts#L193-L196

andrewc5716 (approved):
🚀

PR #54, opened 2025-02-11
fix the sneakiest bug ever

andrewc5716:
i was going nuts until i figured this one out


jbwong05 (approved):
Great find. Is this why we had problems with the peoples endpoint before?

andrewc5716:
i don't think so, that's a separate issue so i was going down the wrong path trying to debug this. but now i got admin privs from john ko as a result lol

PR #56, opened 2025-02-12
Only use WHAT column for strikethrough check

jbwong05:
Only use the WHAT column to determine if the row has strikethrough or not

### DIFF
--- a/src/main/scan.ts
+++ b/src/main/scan.ts
@@ -211,7 +211,7 @@ function isRowStrikethrough(rowData: CellData[]): boolean {
     const whatCell: CellData = rowData[WHAT_COL_INDEX];
     const whatCellRuns: RichTextValue[] = whatCell.richTextValue.getRuns();
     // Incrementally ands all of the strikethrough values together to determine if the entire cell has strikethrough
-    return whatCellRuns.every((run) => run.getTextStyle().isStrikethrough());
+    return whatCellRuns.some((run) => run.getTextStyle().isStrikethrough());
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/scan.ts
```diff
@@ -214,23 +202,16 @@ function isValidEventRow(rowData: CellData[]): boolean {
...
-
-    // Incrementally ands all of the strikethrough values together
-    return nonEmptyCells.every((cellData) => {
-        const runs: RichTextValue[] = cellData.richTextValue.getRuns();
-        return runs.every((run) => run.getTextStyle().isStrikethrough())
-    });
+    const whatCell: CellData = rowData[WHAT_COL_INDEX];
+    const whatCellRuns: RichTextValue[] = whatCell.richTextValue.getRuns();
+    // Incrementally ands all of the strikethrough values together to determine if the entire cell has strikethrough
+    return whatCellRuns.every((run) => run.getTextStyle().isStrikethrough());
```
andrewc5716, line 214: i'm assuming we'd want to replace every with some, right?
```suggestion
    return whatCellRuns.some((run) => run.getTextStyle().isStrikethrough());
```

jbwong05: @andrewc5716 If we want to count the cell as having strikethrough when only part of the What cell has strikethrough then we should use `some`, but if we want to count the cell has having strikethrough when only the entire What cell has strikethrough then it should be `every`. Which do you prefer?

andrewc5716: i see, i thought it was possible for a run to exist if there was another style, like if it was both bold and strikethrough, so there'd be 2 runs in that case. or if part of it was bolded, there'd be a second smaller run? i guess i don't understand exactly how the runs get split up

jbwong05: Hmm actually that's a good point. Let me test that case

jbwong05: Switching to `some` to catch these edge run cases

andrewc5716 (approved):

PR #58, opened 2025-02-19
Check column bounds before adding basecamp link to onestop

jbwong05:
Checks the number of columns in the range before attempting to grab the Google sheets cell to add the Basecamp link into


andrewc5716 (approved):
nice, thanks!

PR #59, opened 2025-02-19
Fix undeleting a row

jbwong05:
Ideally we would be able to delete the row's id from it's metadata when deleting the row. As mentioned in Andrew's comment in the task, this requires somehow saving a reference to the Range which is not trivial (I could not figure out a reliable way to do this given that rows can also shift around).

This is a pretty simple solution that just checks if the row has an id meaning that it was saved in the past, but now the row's saved hash is not present indicating that it was deleted and now has been brought back. In this case, we treat the row as a new row creating new todos, new schedule entries, and overwriting the old id with a new one


andrewc5716 (approved):
nice, glad there's a much simpler solution

PR #60, opened 2025-02-25
Allow addition of hyperlinks to trigger row updates

jbwong05:
This PR updates the toString() function which is used to calculate the row hash to determine if a row has changed to factor in hyperlink and strikethrough data. This allows changes to rows that affect hyperlinks or strikethrough but do not change the text contents of the rows to still trigger row updates.

Note: Updating the row hash algorithm will trigger an update for all existing rows on the next run


andrewc5716 (approved):
nice catch, didn't think about this use case. thanks!

PR #63, opened 2025-03-02
Halt execution on basecamp auth error

jbwong05:
Throws an error to halt execution when Basecamp is unauthorized to prevent any unintended behavior from happening. Also fixes an issue with unpinned dependencies that started causing build errors after Typescript released a new version a couple days ago

Note: I cannot get the unit tests working for this. When the jest mock throws the error, the `instanceOf` check fails for some reason. I hardcoded throwing the error for testing but if someone can figure out the unit tests for this that would be helpful


andrewc5716:
another way to do this without having specific catch scenarios after the basecamp api's is to do a auth check at the beginning of the import onestop method. i think it might be cleaner to throw an error early rather than need to scan thru the whole onestop first before sending the first request that would fail.
though it doesn't hurt to have these catch scenarios too.

jbwong05:
@andrewc5716 Added an auth check to the top of the import function

andrewc5716 (approved):

PR #65, opened 2025-03-12
Team table parsing

jbwong05:
First draft of parsing the team table. The basecamp id fetching logic has been updated to try and fetch the id from the member data read from the team table first. If it can't be fetched it will fall back on the existing id fetching logic from the people endpoint

### DIFF
--- a/src/main/members.ts
+++ b/src/main/members.ts
@@ -6,12 +6,13 @@ import { getCellValues } from "./scan";
 
 const MEMBERS_TAB_NAME: string = "Members";
 const COUPLES_TAB_NAME: string = "Couples";
-const NAME_COLUMN_INDEX: number = 0;
-const GENDER_COLUMN_INDEX: number = 1;
-const MARRIED_COLUMN_INDEX: number = 2;
-const PARENT_COLUMN_INDEX: number = 3;
-const CLASS_COLUMN_INDEX: number = 4;
-const ALTERNATE_NAMES_COLUMN_INDEX: number = 5;
+const NAME_COLUMN_INDEX: number = 28;
+const GENDER_COLUMN_INDEX: number = 4;
+const MARRIED_COLUMN_INDEX: number = 6;
+const PARENT_COLUMN_INDEX: number = 7;
+const CLASS_COLUMN_INDEX: number = 8;
+const ALTERNATE_NAMES_COLUMN_INDEX: number = 29;
+const BASECAMP_ID_COLUMN_INDEX: number = 30;
 const HUSBAND_COLUMN_INDEX: number = 0;
 const WIFE_COLUMN_INDEX: number = 1;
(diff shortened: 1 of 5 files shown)


src/main/members.ts
```diff
@@ -39,8 +42,8 @@ function loadMembersFromOnestop(): { memberMap: MemberMap, alternateNamesMap: Al
     const memberMap: MemberMap = {};
     let alternateNamesMap: AliasMap = {};
 
-    // Start at row 1 to skip the table header row
-    for(let i = 1; i < cellValues.length; i++) {
+    // Start at row 3 to skip the table header row
+    for(let i = 3; i < cellValues.length; i++) {
```
andrewc5716, line 46: ahh magic number 🧙

jbwong05:
Won't actually merge until the Team table is imported from the Master Rotation into the Onestop including everyone's aliases/basecamp ids

andrewc5716 (approved):
approving - and will not have this merged until data is migrated on the onestop

src/main/people.ts
```diff
@@ -46,7 +47,16 @@ function getPeoplePath(): string {
...
-    const normalizedPersonName = normalizePersonName(personName);
+    const normalizedPersonName: string = normalizePersonName(personName);
+
+    // Pull the basecamp id from the member properties if possible
+    if(MEMBER_MAP.hasOwnProperty(normalizedPersonName) && MEMBER_MAP[normalizedPersonName].basecampId !== "") {
+        return MEMBER_MAP[normalizedPersonName].basecampId;
+    }
+
+    // Otherwise fallback on the legacy method that relies on the basecamp people endpoint
+    Logger.log(`Falling back to legacy method to fetch basecamp id for ${normalizedPersonName}`);
```
andrewc5716, line 58: i think this might end up being a noisy log if there's a consistent group of ppl who don't add their basecamp id into the sheet

src/main/people.ts
```diff
@@ -46,7 +47,16 @@ function getPeoplePath(): string {
  * @returns the person's Basecamp id
  */
 export function getPersonId(personName: string): string | undefined {
-    const normalizedPersonName = normalizePersonName(personName);
+    const normalizedPersonName: string = normalizePersonName(personName);
+
+    // Pull the basecamp id from the member properties if possible
+    if(MEMBER_MAP.hasOwnProperty(normalizedPersonName) && MEMBER_MAP[normalizedPersonName].basecampId !== "") {
```
andrewc5716, line 53: we want to log if the normalizedPersonName isn't found in MEMBER_MAP - that means they don't have an alias properly added

andrewc5716: actually if they aren't found in the member map, i think it's ok to short circuit and end this entire execution and log the error, then `return undefined`. in an odd case where their name isn't a registered alias, but it matches with their basecamp name, i don't think we'd want that to keep silently passing. it's better to know upfront it's not a registered alias.

PR #66, opened 2025-03-12
Do not process old tabs

jbwong05:
Filters out tabs with old dates so old tabs are not processed


andrewc5716 (commented):
can we get one small unit test for this? thanks!

andrewc5716 (approved):

PR #68, opened 2025-03-19
Remove noisy log message

jbwong05:
This PR removes the noisy log message for missing basecamp ids for people
The team table parsing PR #65 adds the log message for when the person cannot be found from the Team tab and thus needs to fall back on the legacy way of looking up the basecamp id

### DIFF
--- a/src/main/row.ts
+++ b/src/main/row.ts
@@ -303,13 +303,7 @@ function getBasecampLeadNames(row: Row): string[] {
  * @returns array of Basecamp ids
  */
 function getBasecampIdsFromPersonNameList(personNameList: string[]): string[] {
-    return personNameList.map((name) => {
-        const personId: string | undefined = getPersonId(name);
-        if (personId === undefined) {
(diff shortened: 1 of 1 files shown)


andrewc5716:
@jbwong05 sorry for the scope creep, but i realize now one issue is if ppl put down a bad personId in the team sheet, it would result in a 404. i'm thinking we still do need a set of the valid personId's in the basecamp project, and to double check that what's in the sheet is valid. thoughts?

jbwong05:
@andrewc5716 I see. One question is if we are going to be validate the entered in basecamp id with the one from the people's endpoint, then what is the point of having the id on the sheet if the people endpoint is still the pseudo source of truth?

andrewc5716:
@jbwong05 i'm thinking it can help with:
1. reduce manual work for us to find their id's and add it to the sheet ourselves
2. track who actually added themselves so that leads can know who's using basecamp or not

andrewc5716 (approved):
approving for now; my earlier comments i'll add as an enhancement task for resiliency

PR #69, opened 2025-04-01
fix date bug with month rolling up

andrewc5716:
Bug scenario: When the script is run on the 31st, and the next month doesn't have a 31st, all the dates get bumped up a month.
Explanation: If the current date object has the 31st as the day, and its month gets set to a month without the 31st (e.g. april), it rolls up to the next month (may). When constructing the dates for a Row object, it always uses the current date to start with, which will cause this bug on the 31st 🙃


jbwong05 (approved):
Crazy bug. Great find

PR #70, opened 2025-04-07
Add resilience around parsing tabs with malformed dates

jbwong05:
Adds resilience around parsing tabs with malformed dates in the top left cell. If the date in the top left cell cannot be parsed as a Date object, we will parse the date from the tab name which should be in the format DAY Month/Date

### DIFF
--- a/src/main/scan.ts
+++ b/src/main/scan.ts
@@ -196,7 +196,8 @@ function isDateFromTabNameNextYear(date: Date): boolean {
 /**
  * Retrieves the date for a particular daily tab. For a daily tab, the date is located in the top left cell. Because Google Sheets stores 
  * our Date values in local time and the Google Apps Script API attempts to readjust these Date values assuming they are in UTC, we must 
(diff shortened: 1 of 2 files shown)


andrewc5716 (commented):
does this handle the edge case where current date is the next year due to timezone offsets?

src/main/scan.ts
```diff
@@ -188,20 +188,40 @@ function getColumnData(numCols: number, cellRichTextData: RichTextValue[], cellV
...
-    const utcHoursOffset = getUTCHoursOffset(date);
-    date.setHours(date.getHours() + utcHoursOffset);
 
-    return date;
+    if(date instanceof Date) {
+        const utcHoursOffset = getUTCHoursOffset(date);
+        date.setHours(date.getHours() + utcHoursOffset);
+
+        return date;
+    } else {
```
andrewc5716, line 214: can we add a warning log here that we're calculating a fallback date?

andrewc5716 (approved):

PR #71, opened 2025-06-10
lead/helper validations for new rows

andrewc5716:
when new rows are added to daily tabs, add the lead and helper validation functions.

setting this up locally is janky, you need to have the editor open and call the function `onOpen()` first, then `setupOnChangeTrigger()`, but at least this is a one time thing.

### DIFF
--- a/__test__/members.test.ts
+++ b/__test__/members.test.ts
@@ -7,12 +7,13 @@ global.SpreadsheetApp = SpreadsheetApp;
 import { getRandomlyGeneratedAliasTable, getRandomlyGeneratedMemberMap, getRandomlyGeneratedMemberTable, Mock } from './testUtils';
 import { PersonAliasClashError } from '../src/main/error/personAliasClashError';
 
-const NAME_COLUMN_INDEX: number = 0;
-const GENDER_COLUMN_INDEX: number = 1;
-const MARRIED_COLUMN_INDEX: number = 2;
-const PARENT_COLUMN_INDEX: number = 3;
-const CLASS_COLUMN_INDEX: number = 4;
-const ALTERNATE_NAMES_COLUMN_INDEX: number = 5;
+const NAME_COLUMN_INDEX: number = 28;
+const GENDER_COLUMN_INDEX: number = 4;
+const MARRIED_COLUMN_INDEX: number = 6;
+const PARENT_COLUMN_INDEX: number = 7;
+const CLASS_COLUMN_INDEX: number = 8;
+const ALTERNATE_NAMES_COLUMN_INDEX: number = 29;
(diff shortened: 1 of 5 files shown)


src/main/validation.ts
```diff
@@ -3,12 +3,123 @@ import { MEMBER_MAP } from "./members";
...
+const VALIDATION_HELPER_COLUMN_INDEX = 13; // Column M (1-indexed)
+const VALIDATE_LEAD_FUNCTION_NAME = 'validateLeadCellText';
+const VALIDATE_HELPER_FUNCTION_NAME = 'validateHelperCellText';
+
+/**
+ * Sets up the onChange trigger for the spreadsheet to automatically add validation 
+ * formulas when new rows are added to daily tabs. Replaces any existing trigger with the same name.
+ */
+export function setupOnChangeTrigger(): void {
+    const sheet = SpreadsheetApp.getActive();
```
jbwong05, line 28: Do we need to add this function to all tabs or is it possible to just do the daily ones?

src/main/validation.ts
```diff
@@ -3,12 +3,123 @@ import { MEMBER_MAP } from "./members";
...
+const VALIDATION_LEAD_COLUMN_INDEX = 12; // Column L (1-indexed)
+const VALIDATION_HELPER_COLUMN_INDEX = 13; // Column M (1-indexed)
+const VALIDATE_LEAD_FUNCTION_NAME = 'validateLeadCellText';
+const VALIDATE_HELPER_FUNCTION_NAME = 'validateHelperCellText';
+
+/**
+ * Sets up the onChange trigger for the spreadsheet to automatically add validation 
+ * formulas when new rows are added to daily tabs. Replaces any existing trigger with the same name.
+ */
+export function setupOnChangeTrigger(): void {
```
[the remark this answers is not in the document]
andrewc5716: we call it manually. calling it programatically during onOpen doesn't work for some reason, stackoverflow mentions something about permissions not being set until onOpen is finished being called, so we have to manually call onOpen again in the editor, and then this function

__test__/validation.test.ts
```diff
@@ -409,4 +411,176 @@ jest.mock('../src/main/propertiesService', () => ({
 
         expect(validateLeadCellText("Tech: Kegan")).toBe("");
     });
+});
+
+describe("onSpreadsheetChange", () => {
+    let mockSpreadsheet: any;
+    let mockSheet: any;
+    let mockActiveRange: any;
+    let mockRange: any;
```
jbwong05, line 420: I feel like using `any` kind of defeats the point of using types but all good

andrewc5716: ah yes.. well this is just unit tests 🤷

PR #73, opened 2025-08-09
store ppl and groups in row hash to detect changes

andrewc5716:
whenever ppl change groups, we need a forceUpdate to have it take effect b/c the row hash only saves the row string representation, which doesn't include how groups or other aliases are resolved into individual people.
by saving the full resolution of groups and aliases in the row hash, we can detect all changes automatically.


jbwong05 (approved):
Both manually and vibe reviewed ✅

PR #74, opened 2025-08-10
add auto oauth refresh logic

andrewc5716:
i've been manually refreshing tokens every 2 weeks. turns out there's a way to automatically refresh.

this has been tested manually, but this is also a huge change. maybe the safest path forward is to wait the full 2 weeks for a token to expire and then watch the magic happen.


src/main/basecamp.ts
```diff
@@ -119,20 +120,62 @@ export function logout(): void {
...
 
+/**
+ * Diagnostic function to view OAuth token status for troubleshooting.
+ * Shows essential information about authentication state and automatic refresh capability.
+ * This function is read-only and does not take any action.
+ */
+export function viewAuthStatus(): void {
+    const basecampService: OAuth2 = getUnvalidatedBasecampService();
+    const propertyStore = PropertiesService.getUserProperties();
+    const serviceName = OAUTH_BASECAMP_SERVICE_NAME;
```
jbwong05, line 153: Nit: Missing types

jbwong05 (approved):
Mostly vibed reviewed 🙏

